PR #3, opened 2015-02-05
Fixes for windows

jeroenooms:


evanmiller:
Just FYI if you update to the latest ReadStat code, you should be OK keeping `readstat_rdata.c` in the list of compiled files. LZMA support is now only compiled if `HAVE_LZMA` is defined.

hadley:
Thanks to both of you for all these fixes :)

PR #14, opened 2015-02-11
See if I can get it to install without the .2 R version spec

ehrlinger:
Is there a specific reason to restrict this package to the 3.1.2 R point release? Specifically, can it go to >= 3.1.0? 

It doesn't seem to be related to Imports or Suggests dependencies.


hadley:
Nope, just an oversight

hadley:
Thanks!

PR #16, opened 2015-02-21
fix word omission in README

robertzk:


hadley:
Conflicts with another change :/

PR #48, opened 2015-03-17
clarify development version is on GitHub not CRAN

bquast:
could also be helpful to replace the preceding line with:

if (!require('devtools')) install.packages('devtools')

(from w. chang in rstudio leaflet README.md)


hadley:
Thanks!

PR #102, opened 2015-09-11
Date format in SPSS should return a date and not a datetime

larmarange:


hadley:
It took me a while to understand this too - the format string only affects the display. The underlying data is identical for both dates and date times

PR #103, opened 2015-09-11
Date variables in SPSS should be imported as date, not datetime

larmarange:
NB: dates are stored in seconds in SPSS

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,9 @@
 
 * fixed bug in `zap_labels()` which was leaving labelled vectors unchanged
   instead of leaving unlabelled vectors unchanged. (@markriseley, #69).
+  
+* SPSS's variables whose format is DATE, ADATE, EDATE, JDATE or SDATE are
+  now imported as `Date` variables instead of `POSIXct`.
 
 # haven 0.2.0
 
--- a/src/DfReader.cpp
+++ b/src/DfReader.cpp
@@ -235,6 +235,8 @@ public:
         value /= 1000;
       return value - offset * 86400;
     case HAVEN_DATE:
+      if (type_ == HAVEN_SPSS) // stored in seconds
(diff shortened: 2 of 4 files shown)


hadley:
I'd say that SPSS doesn't have a date data type, so I don't like this implementation because it confuses the type in SPSS with its representation in R.

larmarange:
One of the main issue is that once imported, it's not possible to know what were the original formats in SPSS. Therefore it's not possible to create a function to reconvert from POSIXct to Date only the variables who had a Date format in SPSS (while keeping POSIXct for variables who had a DateTime format in SPSS).

I would argue that it could be therefore an optional argument for `read_spss` (for converting variables with a date format to Date instead of POSIXct, even if I understood that the underlying way to store this information is always datetime in SPSS) OR another option would be to attach an attribute (to the imported data.frame or to the variable) with the content of the original format in SPSS (allowing to write a custom post-import function).

hadley:
I don't think it needs to be an option - I just think your current implantation is confusing. I think the idea is fine

larmarange:
> I think the idea is fine

Sorry, I'm confused. Which idea are referring to?

hadley:
The basic idea of the PR is fine (turn them into dates) is fine, but I don't like the implementation, because I think it's confusing.

larmarange:
Alright. As you have a better knowledge of the code of the package, would you have a suggestion?

hadley:
Hmmmm, I'm warming up to this approach :) It is simple and does the job.  

Would you mind also including a couple of unit tests? And a bullet point in NEWS.

larmarange:
I have updated the pull request.

Regards

larmarange:
Is there any additional issue remaining? Or could it be merges?

hadley:
Thanks!

PR #110, opened 2015-09-28
Relax type checking of numeric objects in labelled()

lionel-:
This fixes a range of errors due to labels and values having mixed integer/double types. One example where this occurs is when importing the ANES datasets in DTA format.

In particular this caused the `[.labelled` method to fail, and also `print.tbl_df()` as a remote consequence.

### DIFF
--- a/R/labelled.R
+++ b/R/labelled.R
@@ -44,7 +44,7 @@ labelled <- function(x, labels, is_na = NULL) {
   if (!is.numeric(x) && !is.character(x)) {
     stop("`x` must be either numeric or a character vector", call. = FALSE)
   }
-  if (typeof(x) != typeof(labels)) {
+  if (!is_coercible(x, labels)) {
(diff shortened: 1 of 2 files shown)


R/labelled.R
```diff
@@ -66,6 +66,18 @@ labelled <- function(x, labels, is_na = NULL) {
   )
 }
 
+is_coercible <- function(x, labels) {
+  if (typeof(x) == typeof(labels)) {
+    return(TRUE)
+  }
+
+  if (all(c(typeof(x), typeof(labels)) %in% c("integer", "double"))) {
```
hadley: Maybe just if `is.numeric(x) && is.numeric(labels)` ?

hadley:
And could you please add a bullet to NEWS?

hadley:
Don't worry, did it myself :)

PR #111, opened 2015-09-30
Factor out and export the labels printer

lionel-:
This extracts a convenience function to print the labels of a labelled vector. This is useful to explore a newly imported dataset.


R/labelled.R
```diff
@@ -81,8 +81,33 @@ print.labelled <- function(x, ...) {
...
+#' @examples
+#' \dontrun{
+#' s1 <- labelled(c("M", "M", "F"), c(Male = "M", Female = "F"))
+#' s2 <- labelled(c(1, 1, 2), c(Male = 1, Female = 2))
+#' labelled_df <- dplyr::data_frame(s1, s2)
+#'
+#' labelled_df %>% purrr::walk(print_labels)
+#' labelled_df %>% purrr::walk2(print_labels, names(.))
+#' }
+print_labels <- function(x, name = NULL) {
```
hadley: This needs to invisibly return it's input (since it's called primarily for its side effect)

R/labelled.R
```diff
@@ -81,8 +81,33 @@ print.labelled <- function(x, ...) {
...
+#' This is a convenience function, useful to explore the variables of
+#' a newly imported dataset.
+#' @param x A labelled vector
+#' @param name The name of the vector (optional)
+#' @export
+#' @examples
+#' \dontrun{
+#' s1 <- labelled(c("M", "M", "F"), c(Male = "M", Female = "F"))
+#' s2 <- labelled(c(1, 1, 2), c(Male = 1, Female = 2))
+#' labelled_df <- dplyr::data_frame(s1, s2)
```
hadley: Can you switch this to tibble? Haven will probably soon import (#154)

R/labelled.R
```diff
@@ -81,8 +81,33 @@ print.labelled <- function(x, ...) {
...
+#'
+#' labelled_df %>% purrr::walk(print_labels)
```
hadley: I wish we didn't have to use `purrr` here :/

R/labelled.R
krlmlr: \dontrun{} shouldn't be necessary then, either.

R/labelled.R
lionel-: do you prefer `lapply/Map()` or maybe for loops in these examples?

hadley: I'd prefer just loops

hadley:
Thanks!

Issue #101, opened 2015-09-10
Error: Failed to parse when importing SAS dataset


nunsuch123:
I have a SAS dataset with about 300 records and about 30 columns. I receive the following error when using read_sas("filename")

"Error: Failed to parse filename: A row in the file was not the expected length."

I am using RStudio version 0.99.465 for Mac. 

This works quite brilliantly on two other sas datasets of similar shape and content. What might be the issue?

hadley:
Can you share the data? That's the easiest way for me to debug

nunsuch123:
I really wish I could share it but it's clinical trial data. It would take me quite a while to recreate it with dummy information and I guess I might remove the problem in question if I do. It has 99 columns, not the 30 I originally thought. I can provide the field specifications (e.g. Field name, type, label, etc) if this is useful. Thank you for your support. 

Lauren

hadley:
It's likely to be something subtle with the formatting of the columns. Unfortunately it's impossible to know without a reproducible example.  Could you try doing a binary search (e.g. removing half the columns each time) to see if you can narrow down where the problem lies?

evanmiller:
@nunsuch123 are you using the latest GitHub version? That error message sometimes appears with the CRAN release but might be fixed in GitHub. If you're using the GitHub version, then any sample file that illustrates the issue would be helpful.

evanmiller:
I believe this issue is a duplicate of https://github.com/WizardMac/ReadStat/issues/35, which has been fixed in ReadStat master.

You can update to the latest ReadStat code with the instructions here: https://github.com/hadley/haven#updating-readstat

A workaround is to save the file without compression.

evanmiller:
Should be fixed by PR #114

Issue #92, opened 2015-07-17
Unable to load big data files


sargdavid:
Hello Hadley,

I'm having difficulties loading a 3 GB .sas7dat file using haven::read_sas. The file opens in SAS but I get the following error in R:

   Error: Failed to parse C:\XX\xx.sas7bdat: Invalid file, or file has unsupported features.

The table has 15 million + rows and 48 columns.

As a work-around, I used your other package - readr. I exported my SAS file as a .CSV, then used read_csv to read it into R. It would be better, of course, if we could import data directly with read_sas. Is there a way to make it work or is it a limitation of this function?

NOTE: read_csv imported the table correctly but with the following warning/problem:

   Warning message:
   8974944 problems parsing 'xx.csv'. See problems(...) for more details.

   Error in withCallingHandlers(expr, warning = function(w) invokeRestart("muffleWarning")) : 
     argument "x" is missing, with no default

I don't know if that would help...

Thank you!

hadley:
It's probably not an issue with the file size, but some feature that the file uses. Can you reproduce the error when saving a smaller subset of the data that you can share?

sargdavid:
Thank you, let me try that and get back to you...

Regards,

Davit

adulbrich:
Dear Hadley,

Following the post from sargdavid, I had similar issues while trying to import large sas7bdat datasets (> 3 GB approximately). I tried several things:

- importing a 2 GB file: works
- importing 4 and 12 GB files: does not work
- importing a small sample of the 12 GB file: works

Here is the Error while running the code:

```R
> system.time(test.big <- read_sas(fileNameOcv))
Error: Failed to parse D:\Data\hh.sas7bdat: Unable to read file.
Timing stopped at: 0 0 0.06
```

And here is the code I used for benchmarking purposes (I love your package by the way, it's even faster than RevoScaleR):

```R
library(haven)
library(RevoScaleR)

############################################################################
##### 1.81 GB (4050025 rows, 57 columns)
fileName <- "D:/Data/bsn.sas7bdat"

### File Import

system.time(test.small <- read_sas(fileName))
# {haven}
# user  system elapsed 
# 283.2     1.5   284.8 
# user  system elapsed 
# 345.98    1.48  347.64

system.time(test.small <- rxImport(fileName))
# {RevoScaleR}
# user  system elapsed 
# 37.78    1.84 1122.17
# user  system elapsed 
# 51.28    1.47 1198.97

############################################################################
[...]

Issue #34, opened 2015-03-03
read_sas cannot load my .sas7bcat file


ghost:
I'm getting the following error when I include my format catalog:

    data<-read_sas("M:/.../alldata.sas7bdat", b7cat = "L:/.../formats.sas7bcat")

    Error: Failed to parse L:\...\formats.sas7bcat: Invalid file, or file has unsupported features.

The data itself loads fine if I don't specify b7cat. If the dataset has the value labels already applied, should they be loaded with only the sas7bdat argument? They seem to not be loading. That is, the data all come out as regular numeric data.

evanmiller:
If you can point me to the sas7bcat file I'll take a look.

AndrewTaylorHFHS:
Unfortunately all of the sas files I work with are HIPPA protected data located on our servers, so I'm not able to do that. In order to have data transfer over to sas with the value labels, do I *need* to reference the catalog file, or am I doing something wrong since when I only reference the sas7bdat file nothing includes the labels?

I apologize for not having a reproducible example; I can close the issue since I am unable to provide one if that would be appropriate (it's why I closed it originally).

evanmiller:
Thanks for the update. @hadley might be able to shed light on the issue of importing into a data frame that already has value labels applied.

hadley:
Yes, you need the catalog file if you want labels.  Could you try creating an example with dummy data?

cullenjd:
I'm having the same problem. When I reference the catalog file using read_sas, I get: 
>"Error: Failed to parse K:\\...\\formats.sas7bcat: Invalid file, or file has unsupported features." 

My data contains personal health information and I don't have SAS to create dummy data. 

I can send the catalog file if that would be any benefit.

evanmiller:
@cullenjd Yes, a sample catalog file that demonstrates the problem would help.

cullenjd:
Here is the sas7bcat file.

cullenjd:
Here's a link to the sas7bcat file:

https://www.dropbox.com/s/o6xwioy72xs4g3i/formats.sas7bcat?dl=0

evanmiller:
Thanks @cullenjd. I believe I've tracked down the issue, but without an accompanying data set I cannot fully test. I've committed code to ReadStat here:

https://github.com/WizardMac/ReadStat/commit/c73249561eebaf648d00e2efb6504fee457b6db5

You can test this code with the instructions here:

https://github.com/hadley/haven#updating-readstat

[...]

PR #114, opened 2015-10-09
Update ReadStat sources #34 #92 #101

evanmiller:
Brings in the latest and greatest bug fixes from ReadStat, including large file support for Windows and AIX. It also addresses the infamous "A row was not the expected length" error with some compressed SAS files.

One new part of the writer API worth mentioning: 

```{C}
void readstat_variable_add_missing_double_value(readstat_variable_t *variable, double value);
void readstat_variable_add_missing_double_range(readstat_variable_t *variable, double lo, double hi);
```

I.e. for SPSS files you can now declare custom missing values and missing ranges.


hadley:
Thanks!

Issue #98, opened 2015-08-20
write_dta not including format


andrewheiss:
Using R 3.2.1, haven 0.2, and Stata 14, `write_dta` does not create browsable data for Stata. All the data is present and usable in Stata, but the columns in a haven-exported file lack a format, so Stata does not display them in the browse window.

Here's a minimal example:

    library(dplyr)
    library(haven)
    library(foreign)

    set.seed(1234)
    example_data <- data_frame(var1 = rnorm(20, 0, 1), 
                               var2 = rnorm(20, 2, 5),
                               var3 = runif(20, -10, 10),
                               var4 = factor(sample(letters, 20)))

    write_dta(example_data, path="with_haven.dta")
    write.dta(example_data, file="with_foreign.dta")

With foreign, everything works well:

<img width="912" alt="with_foreign" src="https://cloud.githubusercontent.com/assets/73663/9389673/4b988786-473a-11e5-9c5b-b370de628d83.png">

With haven, there is no format (`%9.0g`) and all the columns are blank, but the data is still there (see top red box). If I add a format manually, the data displays just fine.

<img width="912" alt="with_haven" src="https://cloud.githubusercontent.com/assets/73663/9389674/4ba60b68-473a-11e5-81a8-0d6fc568bb18.png">

evanmiller:
Fixed in ReadStat: https://github.com/WizardMac/ReadStat/commit/d41fcb6e7f99bf992dc4d4aa8837c8d4f24376c0

(Haven can override the defaults if desired)

PR #118, opened 2015-10-14
Update ReadStat sources #94 #98 #117

### LINKED ISSUE #94: write_dta produces odd output in stata browser
image

image

### LINKED ISSUE #117: Multiple missings in SAS
Using the development version of ReadStat, haven can now handle missing values in value labels from SAS much better than before.However because SAS has multiple values of missing (., .A, .B, etc.), I cannot get all of the data from SAS into R.

Here is a minimal example, the current behavior is pretty good, but the different forms of missing are squashed together, and there are NaNs in the value label.  Given that R only has one value for missing, I can see how this may actually be the best possible way to handle them. However, I thought I'd post this example in case it helps spur a better solution. Possibly it'd be better to drop the value labels with NaNs because they cannot be used and will just be junking up the factor levels.

Thanks!

``` r
library(haven)

data <- read_sas("https://dl.dropboxusercontent.com/u/2019891/haven_bug/formats_missing.sas7bdat", 
                 "https://dl.dropboxusercontent.com/u/2019891/haven_bug/formats_missing.sas7bcat")

data$variable
#> <Labelled>
#> [1]  1  2 NA NA
#> 
#> Labels:
#>  value   label is_na
#>    NaN REFUSED FALSE
#>      1     Yes FALSE
#>      2      No FALSE

as_factor(data$variable, dropNA = FALSE)
#> [1] Yes  No   <NA> <NA>
#> Levels: REFUSED Yes No

as_factor(data$variable)
#> [1] Yes  No   <NA> <NA>
#> Levels: REFUSED Yes No


# desired behavior
# as_factor(data$variable, dropNA = FALSE)
#> [1] Yes  No   REFUSED <NA>
#> Levels: REFUSED Yes No

# as_factor(data$variable)
#> [1] Yes  No   <NA>  <NA>
#> Levels: Yes No
```

evanmiller:
* Export a default format for Stata variables. Fixes #94 #98

 * Support for tagged missing values from SAS and Stata (haven will need to decide what, if anything, to do with these). Addresses but does not fix #117

 * Many improvements to SAS catalog importer


evanmiller:
There appears to be a bug in the SAS catalog importer -- will look into and amend.

hadley:
Thanks!

It would be nice to also maintain a bit of a changelog - it probably doesn't need to be too detailed, but see `NEWS.md` for what I've done in the past.

PR #122, opened 2015-10-19
Update ReadStat sources #121

### LINKED ISSUE #121: SAS catalog: value label sets with long names aren't being matched
SAS catalog files can name their formats (value label sets) with both 8-byte short name and a 32-byte long name. At present ReadStat is returning the short name, which may not be the correct behavior. @cullenjd has reported that his value labels are not being properly applied (see previous discussion at #34). It is likely an issue with ReadStat but could also be something with haven.

evanmiller:
Value label set names longer than 8 bytes (e.g. `CLAVICLE_FORMAT`) were not being correctly matched to their variables. Fixes #121.


hadley:
Thanks!

PR #123, opened 2015-10-21
Add support for variable formats

### LINKED ISSUE #119: Feature request: Add a way of specifying output formats
When writing a .sav file, numeric variables use F8.2 as their format, and string variables default to the length of the longest string in the field.

It would be great to be able to specify variable format, possibly by passing an optional named vector to write_sav.

gorcha:
Feature request (closes #119) - add support for variable formats when reading and writing. Similarly to variable labels, formats are stored as an attribute on each vector. Missing or invalid formats (e.g. trying to assign a string format to a numeric) get the default format for the variable type.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -25,6 +25,8 @@
 * fixed bug in `zap_labels()` which was leaving labelled vectors unchanged
   instead of leaving unlabelled vectors unchanged. (@markriseley, #69).
 
+* added support for writing out variable formats to write_sav and write_dta.
+
 # haven 0.2.0
 
 * fixed a bug in `as_factor.labelled`, which generated <NA>'s and wrong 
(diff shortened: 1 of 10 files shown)


evanmiller:
It is perhaps worth noting that SAV, DTA, and SAS7BDAT formats are incompatible. In your current implementation, a format read from an SAV file and then (unwittingly) written to a DTA file may have unexpected results.

Ideally these formats would be converted to some kind of intermediate representation either in ReadStat or in haven, but this of course will be a lot of work. A workaround would be to store format strings in separate attributes, `format.spss`, `format.stata`, etc.

gorcha:
Thanks Evan, good idea. Have modified so the attributes are FileType specific.

hadley:
I think this is a reasonable idea, but it will need to be rewritten a little to work with #145, which I think is a good approach in the long term.

hadley:
Overall, I like this idea, although it's going to need some updates with the changes I've made the internals. Do you have time to have a stab at updating this PR in the next week or so?

I think it might also be useful to implement a `zap_formats()` function that would drop all the format attributes, which you can then use in the unit tests.

hadley:
This needs quite a bit of work after other changes, so I'll take it on

PR #125, opened 2015-10-22
Handle R connections

### LINKED ISSUE #109: R connections as input
It would be nice to support R connections and automatically handle compressed files, as in readr.

lionel-:
This is a sketch for handling R connections and raw vectors (closes #109), currently only implemented for Stata files.

* It relies on abstracting in a callback API all file input operations in ReadStat.

* A callbacked unistd handling is added in `readstat_io_unistd.c`.

* The DTA parser is updated to use the callbacks. The legacy implementation is currently kept in `readstat_dta_direct.c` for comparison of outputs.

* Connections are transformed to RawVectors through `readr::datasource()` and read with `std::stringstream`

Some benchmarks:
```r
# 1.5 kb file

#> neval: 100
#>         expr median         unit
#> 1     direct     96 microseconds
#> 2  unistd_cb     97 microseconds
#> 3     cpp_cb    275 microseconds
#> 4 gzipped_cb    368 microseconds


# 50 mb file

#> neval: 100
#>        expr median          unit
#>      direct   1968  milliseconds
#>   unistd_cb   2144  milliseconds
#>      cpp_cb   2333  milliseconds
#>  gzipped_cb   2605  milliseconds
```

If you and @evanmiller think this strategy is ok, I'll propose a more complete PR to ReadStat.


evanmiller:
This approach looks good to me. If you're willing to bring it to the rest of ReadStat then I'll work on getting it merged into that project. (We could also do something similar for write handlers, but this seems less important.)

hadley:
Nice!

Does ReadStat need arbitrary file reading? If not, and it can work with a stream of bytes, you could probably avoid a few copies (i.e. you could use the connection API to stream data in). For on disk files, it might be worth considering mmap'd files (a la readr). That said, the performance of haven is already pretty good, so this might not be worth it.

evanmiller:
I had considered mmap, but was concerned about portability esp to Windows so I didn't pursue it. A lot of things will need to be re-written to have a no-copy architecture, so implementing it is a matter of cost/benefit. I'm happy with @lionel-'s approach of just swapping out `read` with `io->read` everywhere, and leaving the door open to future optimizations if copies and syscalls turn out to be a major bottleneck.

lionel-:
> Does ReadStat need arbitrary file reading?

From what I've seen, the DTA backend mostly seeks forward though it checks the end of the file to compute the file size. Maybe that could somehow be avoided. But overall, whether we need random access probably depends on the file format. 

> it might be worth considering mmap'd files

I considered it but I read [here](http://heikok.blogspot.be/2012/08/performance-of-memory-mapped-files-for.html) that the performance is similar. Since the `mapped_file_source` class has methods for `seekg()` and `read()`, it'll be very easy to add a backend. It would be great to collect a set of test files of different sizes so we can make comparisons.

> you could use the connection API to stream data in

That would be nice, and some connections are seekable I think, but unfortunately there's no legal way to get the Rconnection pointer :/

> I'm happy with @lionel-'s approach of just swapping out read with io->read everywhere, leaving the door open to future optimizations if copies and syscalls turn out to be a major bottleneck.

Great! I think there should be no additional copies since everything is passed through pointers, but I was indeed concerned with the calls overhead.

evanmiller:
As you discovered the DTA reader uses seeking mainly to figure out the file size (in order to report back to the progress handler). But the SAS7BDAT reader is a two-pass parser and uses seeking extensively; basically the file format uses text pointers in many places, and the pointed-to text is all sitting at the end of the file. So without seeking, you'd have to read the entire file in and then execute the callbacks all at once at the end of parsing. Many SAS7BDAT files are quite large, so I'd like to stick with the seeking + callback-as-you-go design if possible.

I am unfamiliar with R connections -- is the seeking requirement going to be a problem?

lionel-:
> I am unfamiliar with R connections -- is the seeking requirement going to be a problem?

Not a problem, the streams are read to memory before being passed to readstat. They already work in this PR.

hadley:
Ok, just wanted to check. I think the connection implementation will limit the size of files you can parse (i.e. ~total ram/2) but I doubt that will be a problem in practice.

lionel-:
here you go @hadley

hadley:
@evanmiller would you be interested in incorporating the streaming API into readstat?

evanmiller:
@hadley This work was incorporated here: https://github.com/WizardMac/ReadStat/pull/41

hadley:
Got it - thanks! 

@lionel- do you have time to merge/rebase this PR in the next week? No problems if you don't have time, I'll just do it myself

lionel-:
@hadley No worries, but I'm travelling back home today so I'll look into it tomorrow

lionel-:
@hadley I think this should be good :)

hadley:
Does this mean `clean_path()` can be deleted?

And would you mind adding a few unit tests?

lionel-:
> Does this mean clean_path() can be deleted?

Looks like it.

> And would you mind adding a few unit tests?

oops yes

lionel-:
I added a couple of tests for connections

hadley:
Thanks - much appreciated

PR #135, opened 2015-11-11
Fix missings handling

lionel-:
This is a WIP that currently implements the R side.

I first set out to remove the `is_na` mechanism and replace it with the `missings` attribute we talked about. However I think it will make the analysis less comfortable if we drop the information about which labels correspond to missing values because this is useful to understand a dataset. So this preserves both attributes.

* The user can supply either `is_na`, `missings`, or both. When `is_na` is supplied, we work out the `missings` attribute automatically.

* In the case some pseudo-missing values don't have labels attached, both `is_na` and `missings` will need to be supplied. It's up to the user to make sure that the missingness pattern implied by `missings` is consistent with the one implied by `is_na`. In other words, the `missings` pattern should always be identical or broader than the implied `is_na` pattern. An error is issued when they are not consistent.


lionel-:
This fails because we declare "Male" as missing, but this is not reflected in `missings`:
```r
labelled(
  c("Z", "M", "M", "F", "X", "N/A"),
  labels = c(Male = "M", Female = "F", Refused = "X", "Not applicable" = "N/A"),
  is_na = c(TRUE, FALSE, TRUE, TRUE),
  missings = c(TRUE, FALSE, FALSE, FALSE, TRUE, TRUE)
)
#> Error: Inconsitent missing patterns. Check `missings` and `is_na`
```
This succeeds. The print method shows the number of pseudo-missings:
```r
labelled(
  c("Z", "M", "M", "F", "X", "N/A"),
  labels = c(Male = "M", Female = "F", Refused = "X", "Not applicable" = "N/A"),
  is_na = c(FALSE, FALSE, TRUE, TRUE),
  missings = c(TRUE, FALSE, FALSE, FALSE, TRUE, TRUE)
)
#> [1] Z   M   M   F   X   N/A
#>
#> 3 pseudo-missing values
#>
#> Labels:
#>  value          label is_na
#>      M           Male FALSE
#>      F         Female FALSE
#>      X        Refused  TRUE
#>    N/A Not applicable  TRUE
```

@hadley does this strategy seems reasonable to you?

R/labelled.R
```diff
@@ -13,8 +13,14 @@
 #'   of the values might be labelled.
 #' @param levels When coercing a labelled character vector to a factor, you
 #'   can choose whether to use the labels or the values as the factor levels.
-#' @param is_na Optionally, a logical vector describing which levels should
-#'   be translated to missing values.
+#' @param is_na Optionally, a logical vector that should be as long as
```
hadley: I don't love these names. How about `label_na` and `x_na`?

hadley:
Strategy sounds good to me.

larmarange:
I'm a little afraid that such system would be complex for the final user and not so easy to manipulate.

First of all, storing the same information twice requires continuous checking for consistency and some basic manipulation will maybe become complex.

Taking your example:

```r
v <- labelled(
  c("Z", "M", "M", "F", "X", "N/A"),
  labels = c(Male = "M", Female = "F", Refused = "X", "Not applicable" = "N/A"),
  is_na = c(FALSE, FALSE, TRUE, TRUE),
  missings = c(TRUE, FALSE, FALSE, FALSE, TRUE, TRUE)
)
```

What is happening if I execute `v[1] <- "F"` ?

hat are the rules in case a certain value is sometimes missing sometimes not in the missings attributes ?

What would be the result of 

```r
v2 <- labelled(
  c("Z", "Z", "M", "F", "X", "N/A"),
  labels = c(Male = "M", Female = "F", Refused = "X", "Not applicable" = "N/A"),
  is_na = c(FALSE, FALSE, TRUE, TRUE),
  missings = c(TRUE, FALSE, FALSE, FALSE, TRUE, TRUE)
)
```

where Z is considered as missing in first position, and not in second position ?

How can we easily change which labels are considered as missing? Coming to `v` vector? For any reason, I don't want 'X' to be considered as missing, but I would prefer to consider it as a valid answer. How could I easily modify `v`?

Finally, what would be the pros and cons of such approach compared to the current `is_na` implementation ? compared to the alternative with no `is_na` attribute, and storing in `missings` attributes not a logical vector but the list of values that should be considered as missing (e.g. `c("X", "Z", "N/A")`) ?

Regards

lionel-:
> What is happening if I execute v[1] <- "F"?

We can overload the subset-assignment operators. Also all packages that bypass the subset methods such as dplyr will need to be adapted.

> What would be the result where Z is considered as missing in first position, and not in second position ?

We could add some checks but I'm not sure this would be necessary. I would be fine leaving that responsibility to the user.

> If, for any reason, I don't want 'X' to be considered as missing, but I would prefer to consider it as a valid answer, how could I easily modify v?

Yes we'd need some kind of tool for that. It would allow to easily change the labels and their missingness status, and would reflect the changes in the missings attribute:

```r
na_labels(v) <- c("X", "N/A")
```

> what would be the pros and cons

This approach is simpler because we don't need to address complexities such as ranges of missing values. Furthermore it's really easy to implement thanks to ReadStat missings predicates. However you're right that it does add complexity in other areas, such as subsetting.

Overall, I think the idea is that labelled vectors with pseudo-missings are not first-class citizens, but are rather a convenience and should be transformed to proper R structures as soon as possible. That's why the default for all reader functions is to normalise the pseudo-missings.

Let's see what @hadley thinks.

hadley:
The goal is to create an intermediate data structure that the user can then convert to a pure R object. So I don't think we'd implement subsetting etc - this is just a transitory structure that we don't expect people to use for  long.

lionel-:
@hadley Should I revert e1902f8 then?

hadley:
@lionel- if you've done the work I wouldn't undo, I just would try to avoid going further down that path.

lionel-:
ready of review @hadley 

Travis seems broken right now but R CMD check passes on my machine.

larmarange:
> The goal is to create an intermediate data structure that the user can then convert to a pure R object. So I don't think we'd implement subsetting etc - this is just a transitory structure that we don't expect people to use for long.

I understand your position, in particular for missing values. However, for label values, I don't really agree.  There are clearly two different but similar needs: i) importing data keeping as much information as possible from the external datasets; ii) having a way to deal easily with labelled data in R, not only for import, but also for data manipulation and analysis.

This second need has been visible with several approaches like the data.set format implemented by `memisc`. But this format is not easy to manipulate and not compatible with most R functions/packages.

If we forget for few minutes the issue of missing values, the `labelled` class is very interesting. It's a very nice approach for both need i) and need ii). During the last months, I have been using this class for data manipulation and survey analysis. It allows to store metadata (i.e. variable and value labels) while keeping a true data.frame. It's working well with `dplyr` and `data.table`. For several cases, it's really much easier to manipulate labelled vectors than factors, in particular if we really need to keep the original coding.

----------

Now, it seems that most of problems came when trying to integrate the question of defined missing values. And it seems difficult for this case to find a system allowing both to import with details what was defined in the original dataset and to have something easy to manipulate.

And so far, none of the proposed system was able to deal properly with the `.a` `.b` etc. value in Stata and SAS.

But at least, the labelled attributes (without missing values) are working well and can be used both for import and for data manipulation.

If there is no satisfying system for missing values, would it be possible, at least, to separate clearly attributes for value labels and attributes for missing values, allowing to use them separately?

larmarange:
Sorry for my long messages, it's just that I'm still trying to understand the added value of using a `is_na` attribute and a logical vector, compared to a solution where only the defined missing values would be stored, i.e. something as simple as :

```r
v <- labelled(
  c("Z", "M", "M", "F", "X", "N/A"),
  labels = c(Male = "M", Female = "F", Refused = "X", "Not applicable" = "N/A"),
  missings = c("X", "N/A, "Z")
)
```

From such definition, it is easy to get the logical vector saying which values are missing:

```r
v %in% attr(v, "missings")
```

or to identify which value labels are also missing (for print method for example):

```r
attr(v,"labels") %in% attr(v, "missings")
```

It is something easy to manipulate. No issue with subsetting, assignment or ordering. It is a simple implementation of missings (i.e. no min/max values). There is no consistency to maintain between 'labels' attribute and 'missings' attributes (so much easier to manipulate than the current 'is_na')

For the specific case of SPSS import, using ReadStat missing predicates, it is possible to store (for min/max) all predicted missing values observed in the dataset. On that point, it would be equivalent of your current implementation.

So, why not using such approach?

hadley:
I'm not interested in solving the general dealing with labelled vectors problem. I am only interested in dealing with the how to get labelled vectors from other statistics packages into standard R data structures as easily as possible.

I don't see how your approach deals with missing values defined by ranges. Having a bit vector that stores missing status is the most general approach, and is common in other programming languages (e.g. julia & python).

larmarange:
Sorry for my late answer, things have been complex in Paris.

I understand that your interest is only in terms of importing data.

My concern is mainly from a user point-of-view (not from a coder perspective): once a user import a dataset, there will be cases where he will manipulate his data frame, before transforming/recoding variables. 

A very basic scenario: importing a data frame, converting (in second variable) defined missing values to NA, ordering according to first variable. In the case of a bit vector:

```r
df <- read_spss('mydata.sav')
df$V2[ attr(df$V2, "missings") ] <- NA
df <- df[order(df$V1), ]
```

Let's change the order of the lines:

```r
df <- read_spss('mydata.sav')
df <- df[order(df$V1), ]
df$V2[ attr(df$V2, "missings") ] <- NA
```

The result will be incorrect, because at line 2, when ordering the data frame, the order of the missing attribute doesn't match anymore the order of the data.frame.

For a basic user, it's very difficult to identify such error.

Now, what would be the code if instead of storing a bit vector, we store values who should be considered as missing?

```r
df <- read_spss('mydata.sav')
df$V2[df$V2 %in% attr(df$V2, "missings.values") ] <- NA
df <- df[order(df$V1), ]
```

and if we do:

```r
df <- read_spss('mydata.sav')
df <- df[order(df$V1), ]
df$V2[df$V2 %in% attr(df$V2, "missings.values") ] <- NA
```

The result is identical. From a user point of view, storing values that should be considered as missing instead of a bit vector is easier to use because there are less risk of errors/mistakes.

> I don't see how your approach deals with missing values defined by ranges. 

If in SPSS you have a variable containing 11, 12, 45, 92, 96, 32, 99, 18, 91 and that SPSS define missing values as the range 90 to 100, at import, you can calculate a list of missing values equal to 91, 92, 96, 99, i.e. all observed values that should be considered as missing. Of course, we lose the more general definition of the range of missing values (as with a bit vector), but, at import, information stored is correct.

Concerning the specific issue of importing data from SPSS, both system (bit vector or values that should be considered as missing) are equivalent to describe what was defined in SPSS.

Best regards

lionel-:
> The result will be incorrect, because at line 2, when ordering the data frame, the order of the missing attribute doesn't match anymore the order of the data.frame.

In this case the result will be correct because `[.data.frame` will call the `[.labelled` method which makes sure the missing attribute gets reordered as well. However it's true that if you reorder with dplyr the subsetting method will be bypassed.

Your solution for ranges would work but it could lead to strange results if a variable contains continuous fractional data in the missing range, though that should be an uncommon case.

lionel-:
> Your solution for ranges would work but it could lead to strange results if a variable contains continuous fractional data in the missing range,

Though of course we can just record the integral part.

sjPlot:
If I understand aright, haven would set both missing-attributes: the one that assigns value labels with missing-attributes, and the one that assigns each vector's value with a missing-attribute?

If so, it would or could be task of the user / package to decide which information to use? So simply ignoring one attribute would not matter... I agree with @larmarange that it probably would be a lot of effort to fully support both missing-attributes, especially in terms of re-sorting, deleting values or re-coding (replacing) values.

If it's optional, I would agree with @lionel- initial statement: 

> When is_na is supplied, we work out the missings attribute automatically.

So, whenever the `missings` attribute is needed but missing, simply compute it based on the information from `is_na`.

hadley:
I'm going to close this for now because I don't think it actually solves the problem that we need to solve, namely #170. (My understanding of the problem when we discussed this earlier was sadly lacking). Lets move discussion of the data structure to the issue, and then when we've figured out a good way forward we can probably reuse some of this code.

PR #140, opened 2015-12-01
Separate write tests for sav and dta

diogocp:
These two formats are different enough that they need to be tested
separately. Even if the code paths in haven are similar, some sav
features such as labelled character variables are not supported by the
dta format.

Applying the same tests to write_dta reveals two bugs in the handling of
labelleds.


diogocp:
Superseded by #145.

PR #141, opened 2015-12-01
Additional tests for read_dta (strings and dates)

diogocp:
This adds three new variables to types.dta (string, date and datetime),
in order to test that these types are read correctly.


hadley:
Thanks!

Issue #143, opened 2015-12-02
write_dta should use the smallest possible integral type


diogocp:
Currently all `integer` and `logical` vectors are written as `READSTAT_TYPE_INT32` (Stata's `long`). Stata has two smaller integral types: `byte` and `int` (16 bits). The memory savings from using smaller types can be quite significant when working with large datasets.

write_dta should write all `logical` vectors as `byte` variables, and select the most efficient type for `integer` vectors (by calculating the max and min value).

hadley:
Readstat doesn't currently support the byte type, and in general, I think it's a bad idea to automatically pick the variable type based on the range of the value, as it can lead to confusing behaviour when the values of the variable change.

evanmiller:
For the record, ReadStat does support a byte (char) type (`READSTAT_TYPE_CHAR`).

PR #145, opened 2015-12-04
Fix writing dates/times to dta

### LINKED ISSUE #139: write_dta: dates are off by 10 years
Example:
```
df <- data.frame(date = as.Date("2015-11-30"))
write_dta(df, "date-test.dta")
```
When I open date-test.dta in Stata, the variable has the value 16769. After I apply the %td format, Stata shows it as 29nov2005.

Interestingly, `read_dta` is working correctly. Reading that same file back into R gives me the date 2005-11-29 (as long as I save it in Stata after `format %td`, otherwise it is read as a double).

Stata stores dates as "days since 01jan1960 (01jan1960 = 0)".

### LINKED ISSUE #144: write_dta: numeric labelleds not written correctly
I caught this in #140 but have not yet discovered the cause of the problem.
```
> num <- labelled(c(1, 2), c(a = 1, b = 3))
> num
<Labelled>
[1] 1 2

Labels:
 value label is_na
     1     a FALSE
     3     b FALSE
> roundtrip_var(num)
<Labelled>
[1] 1 2

Labels:
 value label is_na
     0     a FALSE
     0     b FALSE
```

I opened the file in Stata and verified that the labels are not written correctly:
```
. label list
x..i..:
           0 a
           0 b
```

It might be a bug in ReadStat, since the [label-writing code](https://github.com/hadley/haven/blob/master/src/DfWriter.cpp#L165) in DfWriter seems to be basically the same as the code for integer labelleds, which do work fine.

diogocp:
Writing dates and datetimes to dta is currently broken (#139), since haven does not convert them to the Stata epoch (1960-01-01). This pull request fixes that.

Instead of adding more conditionals to write_sav, I opted to create separate write_dta and write_sav methods. I feel that this makes it easier for me and others who are only familiar with one of the formats to contribute. I also expect the two functions to diverge further when #143 is fixed. Perhaps it would make sense at some point in the future to have separate classes for each of the formats inheriting from DfWriter.

Closes #139
Closes #140
Closes #144

### DIFF
--- a/src/DfWriter.cpp
+++ b/src/DfWriter.cpp
@@ -138,7 +138,7 @@ public:
     return Rf_translateCharUTF8(STRING_ELT(label, 0));
   }
 
-  void defineVariable(IntegerVector x, std::string name) {
+  void defineVariable(IntegerVector x, std::string name, const char* format = NULL) {
     readstat_label_set_t* labelSet = NULL;
     if (rClass(x) == "factor") {
       labelSet = readstat_add_label_set(writer_, READSTAT_TYPE_INT32, name.c_str());
@@ -158,10 +158,10 @@ public:
     }
 
     readstat_add_variable(writer_, READSTAT_TYPE_INT32, 0, name.c_str(),
-      var_label(x), NULL, labelSet);
+      var_label(x), format, labelSet);
   }
 
-  void defineVariable(NumericVector x, std::string name) {
+  void defineVariable(NumericVector x, std::string name, const char* format = NULL) {
     readstat_label_set_t* labelSet = NULL;
     if (rClass(x) == "labelled") {
       labelSet = readstat_add_label_set(writer_, READSTAT_TYPE_DOUBLE, name.c_str());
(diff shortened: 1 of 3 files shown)


src/DfWriter.cpp
```diff
@@ -29,6 +29,120 @@ class Writer {
...
+    // Define variables
+    for (int j = 0; j < p; ++j) {
+      RObject col = x_[j];
+      std::string name(names[j]);
+
+      if (col.inherits("Date")) {
+        defineVariable(as<NumericVector>(col), name, "%td");
+      } else if (col.inherits("POSIXct")) {
+        defineVariable(as<NumericVector>(col), name, "%tc");
+      } else switch(TYPEOF(col)) {
```
hadley: Can you use an explicit `{}` block here please?

src/DfWriter.cpp
```diff
@@ -29,6 +29,120 @@ class Writer {
...
+      readstat_begin_row(writer_);
+      for (int j = 0; j < p; ++j) {
+        RObject col = x_[j];
+        readstat_variable_t* var = readstat_get_variable(writer_, j);
+
+        if (col.inherits("Date")) {
+          double val = REAL(col)[i];
+          if (ISNAN(val)) {
+            readstat_insert_missing_value(writer_, var);
+          }
```
hadley: `else` needs to go on same line as `}`

hadley:
My main concern with this approach is that it does lead to a lot of duplicated code, but I think you're right that this is inevitable as the sav and dta writes will continues to diverge overtime.

Could you please complete the split by removing the `type_` field from `Writer`?

hadley:
Actually let me do a little refactoring first

src/DfWriter.cpp
```diff
@@ -29,6 +29,120 @@ class Writer {
...
+          else {
+            val = val + daysOffset(HAVEN_STATA);
+            readstat_insert_double_value(writer_, var, val);
+          }
+        } else if (col.inherits("POSIXct")) {
+          double val = REAL(col)[i];
```
hadley: This needs a bit more thought because `POSIXct`s can be either doubles or integers

hadley:
Done in 944466bdaeb804359930364e3d8bfcc41329c841

PR #154, opened 2016-03-29
Auto-load tibble

krlmlr:
Fixes https://github.com/rstats-db/RSQLite/issues/82 and others.


hadley:
Hmmmm, I've held off on doing this in the past because I didn't want to break the usual data frame behaviour unless you explicitly opted-in by loading dplyr/tibble.

krlmlr:
I'd argue that this is dangerous. If we want opt-in, we should always return a plain data frame and ask users to call as_data_frame(). Otherwise the behavior depends on the load state of a package, which might be not entirely under the control of the user.

How about just declaring this as a potentially breaking change? Users affected by this still can call as.data.frame() on the result.

hadley:
It's under control enough by the user - if they're using dplyr, they'll understand the behaviour. If they're not, they don't need to care about it. This is a big change that will need to happen with readr, readxl, etc in order to be consistent.

krlmlr:
Something else might load the tibble package, without the user being aware of it -- tibble already has reverse imports. Loading tibble is enough to change the behavior, that's why I think the current state is dangerous.

hadley:
Agreed.

Issue #172, opened 2016-06-06
as_factor drops unused labels when levels = "default".


itsdalmo:
As title states, `as_factor.labelled` drops unused labels when `levels = "default"` (or "both").

Example:
```r
  s1 <- labelled(rep(1, 3), c("A" = 1, "B" = 2, "C" = 3))
  exp <- factor(rep("A", 3), levels = c("A", "B", "C"))
  str(as_factor(s1, levels = "default"))
  # Factor w/ 1 level "A": 1 1 1
```

Maybe this is intended, but I'd like to propose that the default should include all labels to safely (and easily) convert labelled variables to factors without information-loss. Users can instead explicitly call `droplevels()` after converting to produce the result above.

-Kristian

hadley:
Can you give an example of where this causes a problem with a realistic labelled example?

itsdalmo:
Started writing this in the pullrequest:

@hadley - My main concern is that there is no option to `levels` which preserves all information when converting a labelled vector to a factor - I.e., you either lose the labels, or you lose the values that are not labelled:

```R
labs <- setNames(1:3, c("A", "B"))
labelled <- labelled(c(1, 3), labels = labs)

as_factor(labelled, levels = "default") # Drops B and C from levels
as_factor(labelled, levels = "labels") # 3 becomes NA
```

However, I see what you mean with your likert example,`as.integer()` on the results would give the wrong values. But perhaps this should only apply when `typeof(x) == "integer"` (or numeric). For character vectors, I think it is better to preserve the order in which the labels were specified in the data, than to sort them alphabetically:

```R
s1 <- labelled(c("N", "A", "D"), c(Agree = "A", Neutral = "N", Disagree = "D"))
as_factor(s1)
#> [1] Neutral  Agree    Disagree
#> Levels: Agree Disagree Neutral
```

E.g., in the example above I think it makes more sense if the outcome is Agree, Neutral, Disagree. And I think this is the best solution in general when the underlying type is `"character"`.

itsdalmo:
To answer your question about a realistic labelled example:
Since the issue outlined here is the default behavior for `as_factor`, I think it's fairly easy to imagine a dataset from a small/medium-sized survey where none of the respondents have ticked a particular label. 

```R
s1 <- labelled(c("N", "A", "D"), c(Agree = "A", Neutral = "N", Disagree = "D", `Don't know` = "U"))
as_factor(s1) # "Don't know" is dropped.
```

[...]

PR #173, opened 2016-06-06
Preserve unused labels when converting labelled with as_factor

itsdalmo:
Feature request and simple fix for #172. The idea is to include all labels to safely convert labelled variables to factors without information-loss. Users can instead explicitly call droplevels() after converting to get the current output.

I also think it would be best to leave labels (in the order they were entered) as the first levels in the resulting factor, and any unexpected values as the last levels (after sorting). e.g:

```r
  s1 <- labelled(c("B", "A", "C"), c("C" = "C"))
  exp <- factor(c("B", "A", "C"), levels = c("C", "A", "B"))
  expect_equal(as_factor(s1), exp)
```

### DIFF
--- a/R/labelled.R
+++ b/R/labelled.R
@@ -152,6 +152,7 @@ as_factor.labelled <- function(x, levels = c("default", "labels", "values", "bot
 
     # Replace each value with its label
     levs <- replace_with(sort(unique(x)), unname(labels), names(labels))
+    levs <- unique(c(names(labels), levs))
     x <- replace_with(x, unname(labels), names(labels))
 
     factor(x, levels = levs, ordered = ordered)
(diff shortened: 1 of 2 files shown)


hadley:
That doesn't seem quite the right behaviour to me. I'm imagining a case like

```R
likert <- labelled(sample(5, 100, replace = TRUE), c("Very Good" = 1, "Very Bad" = 5))
as_factor(likert)
```

i.e. they're just labels - it's the true underlying values that are most important.  If you want different behaviour you can use one of the other options for `levels`.

larmarange:
There are a lot of different cases. This is why having several parameters are useful. 

@itsdalmo in `to_factor` function in `labelled` package (https://cran.r-project.org/web/packages/labelled/) you can specify how you want to sort the levels of the factor.

An `drop_unused_labels` argument is also avalaible in dev version

hadley:
@larmarange yes, and that's why there are four options to `levels`.

hadley:
I'm going to close this for now - lets continue discussion in the issue.

PR #175, opened 2016-06-06
WIP: Basic API for tagged missing values

### LINKED ISSUE #170: Support multiple missing values
SAS and Stata support "tagged" missing values: `.A`, `.B`, ..., `.Z`, and for SAS `._`. SPSS supports per-column user defined missing values: either up to 3 distinct values, or a range plus 1 distinct value. Haven needs to capture these special missing values in a way that preserves regular R missing values semantics, while enabling them to be labelled.

Replaces #33 & #118

hadley:
Fixes #170 

To do:

* [x] ensure labelled class prints tagged missing values specially
* [x] check that `as_factor()` can re-label tagged missings
* [ ] update vignette with examples (including `as_factor()`)
* [x] read SAS/Stata tagged missing values
* [ ] read SPSS missing ranges


src/tagged_na.c
```diff
@@ -0,0 +1,130 @@
...
+// TODO: add helper for extracting first character from first element
+// in character vector. Needs to check for NA and zero length strings
+
+// Vectorised wrappers -----------------------------------------------------
+
+SEXP tagged_na_(SEXP x) {
+  if (TYPEOF(x) != STRSXP)
+    Rf_errorcall(R_NilValue, "`x` must be a character vector");
+
+  int n = Rf_length(x);
```
kevinushey: Do you want to support long vectors here?

itsdalmo:
Your last commit referenced the wrong issue I think. #177 is regarding variable label, did you mean #172?

Would you consider moving the sort to the end, so missing labels are sorted in relation to the existing?

E.g. changing to something like:
```R
# Replace each value with its label
vals <- unique(x)
levs <- replace_with(vals, unname(labels), names(labels))
# Ensure all labels are preserved
levs <- sort(c(setNames(vals, levs), labels))
levs <- unique(names(levs))
```

Would make it so:
```R
s1 <- labelled(c(1, 4), c("Agree" = 1, "Neutral" = 2, "Disagree" = 3, "Don't know" = 5))
as_factor(s1) 
# Returns:
#> Levels: Agree Neutral Disagree 4 Don't know
# Instead of:
#>  Levels: Agree 4 Neutral Disagree Don't know
```

hadley:
Closing because it's now so massive that it's impossible to review.

Issue #178, opened 2016-06-07
reading unix sas7bdat through samba


reikoch:
Lastest version of haven on Windows seems to have lost the ability to read unix sas7bdat from files hosted on a samba share. Could it be due to using readr::datasource()?

> xx <- read_sas('//host/share/dm.sas7bdat')
> ReadStat: Unable to read from file (retval = 2): No error (errno = 0)

evanmiller:
@hadley Could this and other Windows errors be on account of ReadStat seeking within the files? A seek that silently fails would produce the kinds of read and parse errors being reported.

hadley:
Yeah - it looks like we're currently missing a check if the seek fails.   The file stream C++ api is pretty nasty - it doesn't signal failure with either a return value or an exception, but instead sets an "internal state flag"

@lionel- do you have any thoughts?

lionel-:
I guess we should check this flag in the seek methods of the DfReaderInput classes and return an error value if seek failed.

@evanmiller Does ReadStat falls back to reading the stream linearly when seek fails? That may be useful to avoid a copy when reading remote files.

evanmiller:
@lionel- If a seek fails, ReadStat returns with an error (`READSTAT_ERROR_SEEK`).

lionel-:
@reikoch Do the readr functions such as `readr::read_csv()` also fail with Samba shares?

reikoch:
@lionel- readr::read_csv() from the same samba share work just fine.

hadley:
Can you please try again with the dev version?

reikoch:
It works on both PC and Linux, thanks lots! Speed is not that great though, on PC via samba 1GB in 25 minutes, on Linux the same file via NFS in 6 minutes.

lionel-:
Is the speed issue only with dev haven?

reikoch:
Hm, with the dev version from 2016-05-26 the same file via Samba from PC is read in 18 minutes, probably not too big a difference.

PR #180, opened 2016-06-08
Signal failed seek to ReadStat

lionel-:
In the end I did not use [ios::exceptions](http://www.cplusplus.com/reference/ios/ios/exceptions/) because it needlessly makes the code more complex.

Hopefully solves #178


hadley:
Thanks!

PR #188, opened 2016-06-22
Printing also the type of labelled objects

larmarange:
Proposition to also display the type of a labelled object (as this information is "hidden" by the labelled class).

Example:
```
> s1 <- labelled(c("M", "M", "F"), c(Male = "M", Female = "F"))
> s1
<Labelled (character)>
[1] M M F

Labels:
 value  label
     M   Male
     F Female
> s2 <- labelled(c(1, 1, 2), c(Male = 1, Female = 2))
> s2
<Labelled (double)>
[1] 1 1 2

Labels:
 value  label
     1   Male
     2 Female
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -103,6 +103,8 @@
   (@markriseley, #69), and only drops label attributes; it no longer replaces
   labelled values with `NA`s. It also gains a data frame method that zaps
   the labels from every column.
+  
+* `print.labelled()` and `print.labelled_spss()` now display the type.
 
 # haven 0.2.0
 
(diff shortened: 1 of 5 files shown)


larmarange:
In addition, would it be appropriate to also display the variable label?

I.e. something like:

```
<Labelled (character)> Sex of the respondant
[1] M M F

Labels:
 value  label
     M   Male
     F Female
```

hadley:
I don't think the parens are necessary here. 

I'd rather not print the label, I think because it would be more inconsistent with print method for other columns. But I don't feel strongly about it.

larmarange:
The pull request has been updated accordindly

hadley:
Thanks!

PR #203, opened 2016-07-28
Adapt validate_dta to comply with stata variable name length

aghaynes:
Variables names in stata cannot exceed 32 characters. Adapted the REGEX that checks name format to comply with this.


hadley:
Thanks!

aghaynes:
You're welcome. Another patch coming shortly to complete the validation format... :)

PR #205, opened 2016-07-28
Complete validation of stata variable names

aghaynes:
Variable start with either a letter (large or small) or an underscore. Remaining 31 characters can be large or small letter, number or underscore.


hadley:
Thanks!

PR #206, opened 2016-08-05
Move ReadStat library to subdirectory

krlmlr:
for better separation from own code.


hadley:
ReadStat now uses subdirs so I updated MakeVars in an incompatible way.

PR #218, opened 2016-09-13
Added support for SPSS's display widths

ecortens:
In SPSS, columns have a display width attribute that determines the width of the column in the SPSS data file viewer. It does not affect the underlying representation of the data. This pull request adds support for reading and writing display widths in haven, using column-level attributes in a manner similar to the way formats are handled.


ecortens:
Didn't pass build tests, removing.

PR #220, opened 2016-09-23
Add SPSS's display widths to haven

ecortens:
This pull request adds the ability for haven to read and write SPSS's display widths. This field controls only how wide the field appears in the SPSS software and does not affect data storage at all. According to the readstat library, this field is irrelevant for SAS and STATA data files. Display widths are written if the appropriate attribute ("display_width") is set for a given column. The default in Readstat is a display width of 8, and so only non-8 values are preserved as attributes when reading a .sav file. This minimizes the impact on those who are not using display widths. Commit acd4e2d adjusts the .sav file used for testing to have the default display widths, so that they are ignored.


ecortens:
The Travis CI build on the devel branch failed, but it was due to an error installing R on the build system, nothing to do with this pull request. The other two test branches in Travis CI passed.

ecortens:
@hadley I'm just wondering if there's anything I can do to help get this pull request integrated? It's quite a small adjustment, just a few lines, and is using the readstat library, so there shouldn't be any issues with breaking the file format.

hadley:
I'll look at it seriously when I'm next working on a haven release, probably in a month or so to get the new SAS xpt support up on CRAN.

ecortens:
Sounds great!

src/DfWriter.cpp
```diff
@@ -34,6 +34,14 @@ inline readstat_measure_e measureType(SEXP x) {
   }
 }
 
+inline int displayWidth(RObject x) {
+  RObject display_width_obj = x.attr("display_width");
+  if ((display_width_obj.sexp_type() == REALSXP) || (display_width_obj.sexp_type() == INTSXP)) {
+    return (as<IntegerVector>(display_width_obj))[0];
```
hadley: I think you can just return `display_with_obj[0]` here

src/DfWriter.cpp
```diff
@@ -34,6 +34,14 @@ inline readstat_measure_e measureType(SEXP x) {
   }
 }
 
+inline int displayWidth(RObject x) {
+  RObject display_width_obj = x.attr("display_width");
+  if ((display_width_obj.sexp_type() == REALSXP) || (display_width_obj.sexp_type() == INTSXP)) {
```
hadley: Please use `TYPEOF()` to be consistent with the code above.

ecortens: Done!

src/DfWriter.cpp
ecortens: I'll admit, I'm a little unfamiliar with RCpp, but this change results in a compiler error:

src/DfWriter.cpp

Line 40 invalid use of incomplete type 'struct SEXPREC'
Line 40 cannot convert 'SEXPREC' to 'int' in return

I'm unsure how else to convert the generic object to an integer except by casting it to IntegerVector.

hadley: Oops, yes. You need to switch based on the `TYPEOF()` and then use the appropriate helper `INTEGER()` or `REAL()` to extract the first value.

ecortens: Ah, thanks! That's fixed now. In the event of a REAL, it'll be cast to int by the return, truncating the fractional portion, which I think is fine. It's of course not possible to have a fractional display width in SPSS.

hadley:
Thanks!

PR #236, opened 2016-11-04
Fix read_spss documentation & update README.md

pkq:
Fixes reference to Stata in read_spss docs and updates README.md for typos and consistency.

PR #240, opened 2016-11-18
Typo in description of write_sav()

cimentadaj:


pkq:
FYI, this is also fixed in #236

Issue #235, opened 2016-11-02
`write_dta` fails with error (no file written)


g3o2:
The code 
```R
path <- system.file("examples", "iris.dta", package = "haven")
haven::read_dta(path)

tmp <- tempfile(fileext = ".dta")
haven::write_dta(mtcars, tmp)
```

produces the error
```R
Error in eval(substitute(expr), envir, enclos) : 
  Writing failure: Unable to write data.
```

If I replace `write_dta` by `write_sav` or `write_sas`, no error is produced and the file appears to be saved.

```R
> sessionInfo()
R version 3.2.3 (2015-12-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

locale:
[1] LC_COLLATE=French_Luxembourg.1252  LC_CTYPE=French_Luxembourg.1252    LC_MONETARY=French_Luxembourg.1252
[4] LC_NUMERIC=C                       LC_TIME=French_Luxembourg.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] readr_1.0.0    assertthat_0.1 tools_3.2.3    haven_1.0.0    tibble_1.2     Rcpp_0.12.7   
````

This has occurred since haven version 1. Before, the said code would run without any error.

rogerjdeangelis:
Write_sas does not produce an error for me in SAS, however neither WPS nor SAS can read the dataset.

Are you able to read the R SAS dataset in WPS or SAS?

Initially I got the following error from SAS

'dataset is not a SAS dataset'

I slugged in the datset name, which seemed to be missing and I still got a less troublesome error

'I/O error on SAS dataset TEST'

evanmiller:
@rogerjdeangelis Definitely looks like progress. I'll add code to provide a default data set name, so then we can worry about the I/O error.

evanmiller:
See https://github.com/WizardMac/ReadStat/commit/109a629cb5dd27ed4fd89de87223417a6b270db3

g3o2:
@evenmiller and @rogerjdeangelis, in what way are your remarks related to the issue I've created ?

rogerjdeangelis:
Not related to datastep name, but I have examined WPS, SAS  and Stat
Transfer datasets and none of then have the contents of byte 32 = 33 with
the contents of byte(35)=22. I think it might ne better if both bytes
[...]

PR #255, opened 2016-12-23
fix #245 by making sure string columns always have a max_length >= 1

### LINKED ISSUE #245: write_sav() has endless loop, write a 40GB file and still writting
[haven_test.Rdata.zip](https://github.com/tidyverse/haven/files/621619/haven_test.Rdata.zip)

When I use write_sav() to create a SPSS file "test.sav", the file size is growing more than 40GB and still writting.

My Environment:
haven 1.0.0

ecortens:
In https://github.com/WizardMac/ReadStat/commit/39793f9b0dcda35b47fb4796e4553e618c93125d which was integrated into haven at https://github.com/tidyverse/haven/commit/891bccaf314e725d1eac96475b6c5e363ff23479 new code was created to handle the writing of very large strings. However, in so doing, the line that automatically set string columns with width 0 to have a width of 255 was removed. When ReadStat tries to write a string column with width 0, this creates an infinite loop condition on this line https://github.com/WizardMac/ReadStat/blob/master/src/spss/readstat_sav_write.c#L210 since extra_fields starts at -1 and keeps decrementing, and therefore will never be 0, stopping the loop. One fix to this is in the present pull request, which simply makes the minimum length of a string column in haven 1, rather than 0. 

Another fix would be to ReadStat itself, at https://github.com/WizardMac/ReadStat/blob/master/src/spss/readstat_sav_write.c#L903 , to reintroduce the code that overrides a 0 width to 255 as it used to do here https://github.com/WizardMac/ReadStat/blob/085e04876b190a5b7f1d6ad6824c0fa8e74689b6/src/spss/readstat_sav_write.c#L748

I'm unsure which approach is the best, but fixing this issue in haven can't hurt.

### DIFF
--- a/src/DfWriter.cpp
+++ b/src/DfWriter.cpp
@@ -248,7 +248,7 @@ public:
         readstat_label_string_value(labelSet, string_utf8(values, i), string_utf8(labels, i));
     }
 
-    int max_length = 0;
+    int max_length = 1;
     for (int i = 0; i < x.size(); ++i) {
       int length = std::string(x[i]).size();
       if (length > max_length)
--- a/src/readstat/sas/readstat_sas7bdat_write.c
+++ b/src/readstat/sas/readstat_sas7bdat_write.c
@@ -144,7 +144,18 @@ static readstat_error_t sas7bdat_emit_header(readstat_writer_t *writer, sas_head
     };
 
     memcpy(&header_start.magic, sas7bdat_magic_number, sizeof(header_start.magic));
-    strncpy(header_start.file_label, writer->file_label, sizeof(header_start.file_label));
+
+    memset(header_start.file_label, ' ', sizeof(header_start.file_label));
+
+    size_t file_label_len = strlen(writer->file_label);
+    if (file_label_len > sizeof(header_start.file_label))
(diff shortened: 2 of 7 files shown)


ecortens:
Now also update readstat to https://github.com/WizardMac/ReadStat/commit/c443d7f0ba297a083a254bc8257a041616a7039a which fixes both #235 and #245

hadley:
For NEWS, do you know why the update fixed #235?

ecortens:
Sorry for being unclear. This only fixes #235 by updating readstat. The change to the width of string columns only addresses #245.

hadley:
Oh can you please do that clarify that in a separate PR? (And fix the news bullet)

ecortens:
Ah, perhaps I'm misunderstanding. Your commit 42c8883, which updates ReadStat, fixes both #235 and #245. The commit e73a903 in this pull request only addresses #245; it makes sure that ReadStat will never be passed a string column with a width less than 1. But that fix isn't necessary with the update to ReadStat, since it will automatically set a string column of width 0 to 8. So all this to say, you can close/ignore this pull request if you like.

Issue #299, opened 2017-07-16
undefined symbol: libiconv


mfansler:
## Problem

**Ran:** `install.packages("tidyverse")`

**Result:** **haven** compiles, but during the step to test package loading it yields

    ...
     ** testing if installed package can be loaded
     Error in dyn.load(file, DLLpath = DLLpath, ...) :
       unable to load shared object '/home/username/.local/lib/R-3.3.3/haven/libs/haven.so':
       /home/username/.local/lib/R-3.3.3/haven/libs/haven.so: undefined symbol: libiconv
     Error: loading failed
     Execution halted
     ERROR: loading failed
     * removing ‘/home/username/.local/lib/R-3.3.3/haven’

which results in failed installation.

## System Info

    > sessionInfo()
    R version 3.3.3 (2017-03-06)
    Platform: x86_64-pc-linux-gnu (64-bit)
    Running under: CentOS Linux 7 (Core)
    
    locale:
     [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C
     [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8
     [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8
     [7] LC_PAPER=en_US.UTF-8       LC_NAME=C
     [9] LC_ADDRESS=C               LC_TELEPHONE=C
    [11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C
    
    attached base packages:
    [1] stats     graphics  grDevices utils     datasets  methods   base

This is on a cluster account (no root privileges).  Non-R libraries that I require, such as `libiconv` are built and installed into a local path, and that path is included in `LD_LIBRARY_PATH`.  I have verified that `Sys.getenv("LD_LIBRARY_PATH")` includes the path to **libiconv.so**.

## Expected behavior
R should use `LD_LIBRARY_PATH` environment variable to resolve linked libraries, in this case `libiconv`.

## Attempted workarounds

1. Manually pre-load `libiconv`

       dyn.load("/home/username/.local/lib/libiconv.so")
       is.loaded("libiconv") ## [1] TRUE
       install.packages("haven")

2. Create symlink to **libiconv.so** in R library path.

Both attempts resulted in the same error message being encountered when the installer tests loading.

## Additional notes
[...]

PR #300, opened 2017-07-19
Adds explicit link to libiconv

mfansler:
Some linux distributions do not include `iconv` installed by default and therefore require the `libiconv` library to be explicitly linked to the package.  This PR makes that dependency explicit.

More details for the error one encounters otherwise can be found in Issue #299.


mfansler:
Hmm...looks like some systems that have `iconv` defined in `libc` don't have `libiconv`.  Well, at least Travis caught it.

PR #318, opened 2017-10-13
styling with styler

lorenzwalthert:
Styled with styler (and a few manual tweaks). Not a lot to style here though 😄. I applied different levels of invasiveness sequentially. Some cases might be debatable so let me know what you think.


hadley:
Looks good - thanks!

PR #352, opened 2018-02-20
Fix for converting to factor using option "labelled" when there are tagged nas

gergness:
Split out from #351 

In the current version of haven, `as_factor()` gets confused by tagged NAs when using the labels option.
``` r
library(haven)
x1 <- labelled(c(1, 2, tagged_na("a")), c(a = 1, b = tagged_na("a")))
as_factor(x1, "labels")
#> Error in factor(x, levs, labels = labs, ordered = ordered): invalid 'labels'; length 2 should be 1 or 1
```

After PR:
``` r
library(haven)
x1 <- labelled(c(1, 2, tagged_na("a")), c(a = 1, b = tagged_na("a")))
as_factor(x1, "labels")
#> [1] a    <NA> b   
#> Levels: a b
```

PR #351, opened 2018-02-18
Add pillar printing (+ 2 other fixes along the way) Fixes #340

gergness:
Here's an attempt to bring pillar printing to the labelled class. I also encountered 2 bugs along the way so have included fixes for them here. Let me know if you'd prefer that I split them up.

Bug 1) In the current version of haven, `as_factor()` gets confused by tagged NAs when using the labels option.
``` r
library(haven)
x1 <- labelled(c(1, 2, tagged_na("a")), c(a = 1, b = tagged_na("a")))
as_factor(x1, "labels")
#> Error in factor(x, levs, labels = labs, ordered = ordered): invalid 'labels'; length 2 should be 1 or 1
```

Bug 2) In the current version of haven, subsetting labelled_spss objects converts them to labelled class, losing their NA label attributes.
``` r
library(haven)
x2 <- labelled_spss(c(1, 2, 9), c(a = 1), na_values = 9)
is.na(x2)
#> [1] FALSE FALSE  TRUE
is.na(x2[1:3])
#> [1] FALSE FALSE FALSE
class(x2[1:3])
#> [1] "labelled"
```

And here's some examples of the new printing in action:
```{r}
library(haven)
examples <- dplyr::data_frame(
  x1 = labelled_spss(c(1:10, NA), c(Good = 1, Bad = 8), c(9, 10)), 
  x2 = labelled_spss(c(1:10, NA), c(Good = 1, Bad = 8, Refused = 10), c(9, 10)),
  x3 = labelled(c(1:8, tagged_na("a"), tagged_na("b"), NA), c(Good = 1, "Bad" = 8)),
  x4 = labelled(c(1:8, tagged_na("a"), tagged_na("b"), NA), c(Good = 1, "Bad" = 8, Refused = tagged_na("b"))),
  x5 = labelled(1:11, c("Not used" = 12)),
  x6 = labelled(c(rep(c(1.22352, 1000, -345), each = 3), 35, 35), c(One = 1.22352, Two =  35, Threeeee = 1000)),
  x7 = labelled(c(rep("A", 3), rep("B", 3), rep("XXXXXX", 4), NA), c(Apple = "A", Banana = "B", Mystery = "XXXXXX"))
)

examples
```
screen shot 2018-02-18 at 4 32 22 pm

```{r}
print(examples, width = 70)
```
screen shot 2018-02-18 at 4 32 39 pm

```{r}
print(examples, width = 75)
```
screen shot 2018-02-18 at 4 32 52 pm

```{r}
[...]

PR #353, opened 2018-02-20
Preserve labelled_spss class when subsetting

gergness:
Hm, I'm at the edge of my git comfort level, so I'm not sure I've split this up in the best way. Is there a way to have it show the diffs only relative to my previous PR? 

This is the second part of #351, and includes #352. 

Before this PR:
Subsetting labelled_spss objects converts them to labelled class, losing their NA label attributes.
``` r
library(haven)
x2 <- labelled_spss(c(1, 2, 9), c(a = 1), na_values = 9)
is.na(x2)
#> [1] FALSE FALSE  TRUE
is.na(x2[1:3])
#> [1] FALSE FALSE FALSE
class(x2[1:3])
#> [1] "labelled"
```

After PR:
``` r
library(haven)
x2 <- labelled_spss(c(1, 2, 9), c(a = 1), na_values = 9)
is.na(x2)
#> [1] FALSE FALSE  TRUE
is.na(x2[1:3])
#> [1] FALSE FALSE  TRUE
class(x2[1:3])
#> [1] "labelled_spss" "labelled"
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,9 @@
   * Fixes out of memory error (#342)
   * Now supports reading and writing stata 15 files (#339)
 
+* Subsetting `labelled_spss` values keeps the original class instead of converting
+  to `labelled` (#340 @gergness)
+
 * Fix for `as_factor()` when the option levels="labels" and there are tagged NAs
   (#340 @gergness)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,9 @@
   * Fixes out of memory error (#342)
   * Now supports reading and writing stata 15 files (#339)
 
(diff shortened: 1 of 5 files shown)


huftis:
@gergness I stumbled over the same problem before seeing this PR, and filed an issue. :) So you could add that it fixes #360 too. And in your `testthat` example, perhaps you could

```
expect_identical(x, x[1:5])
```

too? This will ensure that any attribute (e.g. `na_values`) is kept when subsetting.

gergness:
Good call, thanks @huftis !

hadley:
Yeah, it would be helpful if you could cleanly tease apart the two sets of changes. I'm not sure how to do this retrospectively, but in the future you can avoid the problem by creating a branch for each PR.

PR #384, opened 2018-06-20
`labelled()` and `labelled_spss()` now support the `label` attribute

### LINKED ISSUE #362: Generators `labelled()` and `labelled_spss()` don’t support adding `label` attribute
The `labelled()` and `labelled_spss()` generator functions supports adding a `labels` attribute, but they don’t have a built-in mechanism for adding a `label` attribute. (The `label`attribute hold a short descripton of the variable.) This despite the fact that `haven` actually has very good support for the `label` attribute, e.g. when both writing and reading SPSS files.

Suggestion: Add a `label` argument to the `labelled()` and `labelled_spss()` functions. This would add the argument content as a `label` attribute. Although the argument logically belongs next to the `labels` argument, it’s probably best to add it as the *last* argument of the functions, in case people have used argument position when calling the functions.

Here’s a reprex illustrating that 1) the generator functions have no arguments for specifying the `label` attribute, and 2) exporting and importing to the SPSS format *does* support the `label` attribute (and correctly loses ‘unknown’ attributes):

```r
library(haven)
library(tibble)

# No label argument
args(labelled)
#> function (x, labels) 
#> NULL
args(labelled_spss)
#> function (x, labels, na_values = NULL, na_range = NULL) 
#> NULL

# Example data
gen = labelled(c(0, 0, 1, 99), 
               labels = c(Female = 0, Male = 1))
attr(gen, "label") = "Patient gender"
attr(gen, "foo") = "bar"
d = tibble(gen)

# Write to SPSS and read resulting file
fname = tempfile()
write_sav(d, fname)
d2 = read_sav(fname)

# The `label` and `labels` attributes survived the conversion,
# but the `foo` attribute (correctly) did not
str(d2)
#> Classes 'tbl_df', 'tbl' and 'data.frame':    4 obs. of  1 variable:
#>  $ gen:Class 'labelled'  atomic [1:4] 0 0 1 99
#>   .. ..- attr(*, "label")= chr "Patient gender"
#>   .. ..- attr(*, "format.spss")= chr "F8.2"
#>   .. ..- attr(*, "labels")= Named num [1:2] 0 1
#>   .. .. ..- attr(*, "names")= chr [1:2] "Female" "Male"
```

I could probably create a PR if this is of interest.

huftis:
The constructor functions `labelled()` and `labelled_spss()` now support
adding the `label` attribute to the resulting object. The `label` is a
short, human-readable description of the object, and is now also used
when printing, and can be easily removed using the new `zap_label()`
function. Fixes #362.

Previously, the `label` attribute was supported both when reading
and writing SPSS files, but it was not possible to actually create
objects in R having the `label` attribute using the constructors
`labelled()` or `labelled_spss()`.

Note that I have not added the automatically generated
documentation files (`.Rd` files) in this commit. I wasn’t sure
if it was best to to add them or just let them be automatically
generated sometime in the future (when releasing the next
version of the package).


hadley:
Can you please fix the build failure?

hadley:
And this will probably have to wait until the next release.

hadley:
Thanks!

PR #391, opened 2018-06-29
Updated Readme to advertise support of stata 15

oscardssmith:
The repository already supports stata 15, but the readme didn't explicitly say so.


hadley:
Thanks!

PR #403, opened 2018-09-14
Fix reference to sas7bcat file in man for read_sas

nealrichardson:
I noticed what looks to be a typo. The README seems to reference sas7bcat correctly but it was not in read_sas


hadley:
Thanks!

PR #407, opened 2018-09-26
pass dots from as_factor.data.frame to as_factor.haven_labelled

zkamvar:
Arguments to the `haven_labelled` class were not being passed from the `as_factor.data.frame()` method and were being silently ignored:

``` r
library("haven")
x <- labelled(sample(5, 10, replace = TRUE), c(Bad = 1, Good = 5))
y <- data.frame(list(x = x))
as_factor(x, levels = "values")
#>  [1] <NA> 1    5    5    <NA> 1    1    <NA> 1    5   
#> Levels: 1 5
as_factor(y, levels = "values")$x # should be the same as above
#>  [1] 4    Bad  Good Good 4    Bad  Bad  3    Bad  Good
#> Levels: Bad 3 4 Good
```

Created on 2018-09-26 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

This PR should fix that


hadley:
Would you mind including a unit test?

Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks! I finished it off.

PR #421, opened 2019-01-19
Replace data_frame with tibble

thays42:
This PR replaces calls of data_frame (deprecated) with tibble, per tidyverse/dplyr#4069 for Tidyverse Dev Day.


hadley:
Thanks!!

PR #425, opened 2019-02-07
Allow non-unique labels when as_factor with levels = 'label'. Fixes #424

### LINKED ISSUE #424: non-unique labels
Similar but not the same as #364 - it's currently possible to have multiple values mapped to a label with the same text.

```r
library(haven)
library(tibble)

x <- labelled(1:2, c("label" = 1, "label" = 2))
x
#> <Labelled integer>
#> [1] 1 2
#> 
#> Labels:
#>  value label
#>      1 label
#>      2 label
```

However, if you do this you can't assign to factors using option `levels = "label"`

```r
as_factor(x, "labels")
#> Error in `levels<-`(`*tmp*`, value = as.character(levels)): factor level [2] is duplicated
```

This is especially problematic because #390 used this in the print method.
```r
tibble(x = x)
#> Error in `levels<-`(`*tmp*`, value = as.character(levels)): factor level [2] is duplicated
```

Stata at least considers this data valid ([test.zip](https://github.com/tidyverse/haven/files/2842301/test.zip) attached), so I think the best solution is to fix the `as_factor()` code, but if not, I think it's important to at least change the tibble print method.

```r
read_dta("~/test.dta")
#> Error in `levels<-`(`*tmp*`, value = as.character(levels)): factor level [2] is duplicated
```

gergness:
Fixes #424


gergness:
the build failure seems unrelated: 
> The command "curl -fLo /tmp/pandoc-2.2-1-amd64.deb https://github.com/jgm/pandoc/releases/download/2.2/pandoc-2.2-1-amd64.deb" failed and exited with 56 during .

hadley:
Thanks!!

PR #426, opened 2019-02-12
Fix NEWS items

batpigandme:
* Change "used" to "uses".
* Clarify the column-naming bullet.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,7 +30,7 @@
     * `write_sav()` no longer creates incorrect column names when >10k columns (#410)
     * `write_sav()` no longer crashes when writing long label names (#395)
   
-# haven 2.0.0R
+# haven 2.0.0
 
 ## BREAKING CHANGES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,7 +22,7 @@
 *   Updated to latest ReadStat from @evanmiller:
 
     * `read_por()` can now read files from SPSS 25 (#412)
-    * `read_por()` uses base-10 instead of base-30 for the exponent (#413)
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -22,12 +22,12 @@
 *   Updated to latest ReadStat from @evanmiller:
 
     * `read_por()` can now read files from SPSS 25 (#412)
-    * `read_por()` used base-10 instead of base-30 for the exponent (#413)
+    * `read_por()` uses base-10 instead of base-30 for the exponent (#413)
```
hadley: This one actually was correct - it's _supposed_ to use base 30 🤷‍♂️

batpigandme: Should we go to the convention of:
_now uses base-30 instead of base-10 for the exponent_?

hadley: Oh yes, that's much clearer

NEWS.md
```diff
@@ -22,12 +22,12 @@
...
+    * `read_por()` now uses base-30 instead of base-10 for the exponent (#413)
     * `read_sas()` can read zero column file (#420)
     * `read_sav()` reads long strings (#381)
     * `read_sav()` has greater memory limit allowing it to read more labels (#418)
     * `read_spss()` reads long variable labels (#422)
-    * `write_sav()` creates incorrect column names when >10k columns (#410)
+    * `write_sav()` no longer creates incorrect column names when >10k columns (#410)
     * `write_sav()` no longer crashes when writing long label names (#395)
   
 # haven 2.0.0R
```
batpigandme: @hadley, is the R here intentional? If no, I figure might as well get rid of it in this PR.

hadley: I
t's not

PR #439, opened 2019-03-06
Support (IS|E|B)8601(DT|DA|TM) SAS formats

mikmart:
There's a whole bunch of ISO 8601 date/time formats for SAS that could be supported (e.g. [`E8601DA`](http://support.sas.com/documentation/cdl/en/lrdict/64316/HTML/default/viewer.htm#a003065455.htm)) . I added a few that come up quite often.


mikmart:
It might be nice to have a little helper function that could handle all of the formats in this class (there are also  DN, DZ, LZ and TZ suffixed versions of these), rather than adding each as a new case.

mikmart:
If not all of these, could at least support for one of the datetime formats be added? At the moment there is no way to format dates in SAS according to IS0-8601 and still have them read in as dates by haven.

hadley:
I think at some point we should reconsider this code, but just adding individually seems fine for now.

Issue #248, opened 2016-12-01
Import subset of columns


carlganz:
Similar to `readr::cols_only`, it would be very nice if I didn't have to load entire SAS datasets into R just to use one or two variables. Let me know if I can help.

Regards

hadley:
This is a relatively complicated problem because you need to figure out some good way of describing which columns you want to important. It might be possible to borrow readr's `cols()` specification, but that feels like a bit of an abuse to me, because haven doesn't have to do any coercion.

It would be helpful if you could provide some examples of the problems you are trying to solve.

carlganz:
An example is I work in a data access center where data is kept exclusively in (large) SAS files. I develop Shiny apps, which have to read the entire SAS file, which takes a while, even though the app is only using a small subset of the variables in the file. Obviously SQL is the solution, but in the world of public health SAS/Stata/SPSS is ubiquitous, and SQL is rare (for now).

My hope is that I can specify a subset of the columns when I read in the SAS files, and speed up the reading process. Honestly, I haven't looked at the source code enough to know if that is even feasible.

Regards

hadley:
Could you give me an idea of the size of the files? (in terms megabytes, rows x cols, and how long it takes to load?).  Actually importing selected columns will be relatively straightforward, but I'm not sure if it will improve performance by that much.

carlganz:
For a one year Adult [CHIS](http://healthpolicy.ucla.edu/chis/Pages/default.aspx) file you will find around 21000 rows and 2400 columnn, so a little over half a terabyte in size. It takes a little under 20 seconds to load. 

Sometimes I just need one variable from two separate years so I end up loading essentially a whole terabyte of data into R just to work with a really small subset of that data. 

Happy to chat about this with you at CSP next month if you aren't too busy.

hadley:
Ah, I was thinking about speed, but memory is also an issue for that much data. Even if it was only marginally faster, it would still be helpful to pull in only selected columns.

@evanmiller have you thought about this at all? Obviously I could filter based on the variable index, but maybe if haven knew which variables I was interested in it could skip even more work?

evanmiller:
Skipping columns will likely be the same amount of I/O but could cut down on CPU significantly. My guess is the perf gain will depend on whether the workload is I/O-bound or CPU-bound.

[...]

PR #440, opened 2019-03-09
Column selection and row limiting on read

mikmart:
Implements selective column reading (#248) powered by **tidyselect**.

General strategy is to do two reads: one with 0 rows to find the column names to use with `vars_select()`; and a second read for the full file with just the selected columns. The additional read adds a bit of overhead (which also scales with the number of rows in the data), but performs decently compared to a full read (c.f. [this benchmark](https://github.com/tidyverse/haven/issues/248#issuecomment-471053100)).

To facilitate the first read, an `n_max` argument is also implemented (with ReadStat's `row_limit` API). Currently ReadStat doesn't support limiting to 0 rows (used as "all" instead), so 0-row behaviour is faked at the C++ level by just reading 1 row and not returning it when the user asks for `n_max = 0` at the R level.


hadley:
Basic approach looks sound. I think the main thing is to switch to using `col_select` (to match vroom); I think that just involves changing some variable names here and there and then [deprecating `cols_only`](https://github.com/tidyverse/principles/issues/67) in `read_sas()`.

mikmart:
Alright! I think I've handled the deprecation now.

R/haven.R
```diff
@@ -32,12 +37,25 @@ NULL
 #' read_sas(path)
 read_sas <- function(data_file, catalog_file = NULL,
                      encoding = NULL, catalog_encoding = encoding,
-                     cols_only = NULL) {
+                     col_select = NULL, n_max = -1L, cols_only = "DEPRECATED") {
+  if (!missing(cols_only)) {
+    # cols_only never used tidyselect, so should error on bare names
+    col_select <- rlang::quo(!!cols_only)
```
hadley: I think this should be `syms()`?

mikmart:
Absolutely, I'll have a look.

R/haven.R
mikmart: Oh right, yeah. This would work with numeric selections, even though that wasn't possible before.

R/haven.R
```diff
@@ -39,9 +39,11 @@ read_sas <- function(data_file, catalog_file = NULL,
                      encoding = NULL, catalog_encoding = encoding,
                      col_select = NULL, n_max = -1L, cols_only = "DEPRECATED") {
   if (!missing(cols_only)) {
-    # cols_only never used tidyselect, so should error on bare names
-    col_select <- rlang::quo(!!cols_only)
     warning("`cols_only` is deprecated. Please use `col_select` instead.", call. = FALSE)
+    stopifnot(is.character(cols_only)) # used to only work with a char vector
+
+    # guarantee a quosure to keep NULL and tidyselect logic clean downstream
+    col_select <- rlang::quo(c(!!!cols_only))
```
mikmart: Is this OK? I think using `syms()` would make downstream messy as would need to handle both the quosure and list of symbols case. Now using `cols_only = c("foo", "bar")` should be pretty much exactly the same as `col_select = c("foo", "bar")`.

hadley: Yeah, that looks good.

mikmart:
So I ended up refactoring DfReader.cpp quite a lot to avoid copying so much code. If this is too much, I'll roll it back a bit and just copy over the changes from the SAS implementation.

Everything should be working now, though.

R/haven.R
```diff
@@ -249,23 +268,25 @@ read_spss <- function(file, user_na = FALSE) {
 #' write_dta(mtcars, tmp)
 #' read_dta(tmp)
 #' read_stata(tmp)
-read_dta <- function(file, encoding = NULL) {
+read_dta <- function(file, encoding = NULL, col_select = NULL, n_max = -1L) {
   if (is.null(encoding)) {
     encoding <- ""
   }
 
+  cols_skip <- skip_cols(read_dta, {{ col_select }}, file, encoding)
```
hadley: Since you use `{{ }}` can you please require rlang 0.4.0 in the DESCRIPTION?

src/DfReader.cpp
```diff
@@ -115,7 +115,9 @@ class LabelSet {
 // DfReader ------------------------------------------------------------------
 
 class DfReader {
-  FileType type_;
+  FileExt ext_;
+  FileVendor vendor_;
```
hadley: What do you mean by vendor here?

R/haven.R
```diff
@@ -362,3 +383,14 @@ var_names <- function(data, i) {
...
 }
+
+skip_cols <- function(reader, col_select = NULL, ...) {
+  col_select <- rlang::enquo(col_select)
+  if (rlang::quo_is_null(col_select)) {
+    return(character())
+  }
+
+  cols <- names(reader(..., n_max = 0L))
+  sels <- tidyselect::vars_select(cols, !!col_select)
```
hadley: Maybe this should error if `length(sels) == 0`?

src/DfReader.cpp
```diff
@@ -371,6 +377,12 @@ class DfReader {
     }
   }
 
+  void limitRows(long n) {
+    if (n >= 0 && nrows_ > n) {
```
hadley: It feels a bit weird to simply ignore out of range values in an internal setter. Maybe it would be better to do input validation higher up?

src/DfReader.cpp
```diff
@@ -535,150 +559,77 @@ readstat_error_t dfreader_update(long file_size, readstat_progress_handler progr
...
-template<typename InputClass>
-List df_parse_xpt(Rcpp::List spec, std::string encoding = "") {
-  DfReader builder(HAVEN_XPT);
-  InputClass builder_input(spec);
-
-  readstat_parser_t* parser = haven_init_parser(encoding);
+template<FileExt ext>
+void haven_parse(readstat_parser_t* parser, DfReaderInput& builder_input, DfReader* builder) {
   haven_init_io(parser, builder_input);
+  readstat_error_t result = [&] {
```
hadley: Can you avoid the use of an anonymous function here?

src/DfReader.cpp
mikmart: Vendor is meant to refer to SAS/Stata/SPSS, for determining formats etc. I considered just adding the "ext" and keeping "type", but I thought "vendor" could be more expressive.

src/DfReader.cpp
mikmart: I moved the `n >= 0` check out of the setter in 3e6c69018dc5e07df125c11014c8420dc17cf853, but now it feels weird to possibly allow setting `n_rows` to a negative value 🙈

Currently I've documented `n_max = -1` as "all rows" but silently all negative values are used for that. Do you think this should be stricter?

mikmart:
Yeah I was a bit conscious about the change bloat with the refactoring.. So the better approach would have been to submit a separate PR for that part?

hadley:
Yeah, generally, smaller PRs are easier for me to review — the challenge is that often don't know that you need a refactoring until you've started on the feature work.

hadley:
Thanks for all your work on this!

Issue #444, opened 2019-04-13
Add look_for_* search functions?


briatte:
In the spirit of #237 and #443, would there be an interest in offering `haven` users functions to search through variable labels or values, which are very much needed for interactive exploration of labelled data such as large survey datasets?

This would involve:

1. Allowing the `print_labels` and my `print_variable_labels` suggestion in #443 to return their results completely invisibly
2. Writing two functions, `look_for_labels` and `look_for_variables`, that would subset the corresponding `print_*` results

Step 2 can be done, using `grepl` as in #237, although it'd have to grep both variable/value names _and_ labels and support at least `ignore.case`, or it could use `tidyselect`, as in

```r
look_for_variables(labelled_df, starts_with("Municipal"))
look_for_labels(labelled_var, contains("employed"))
```

Another verb for the functions could be `find_*`.

## Rationale

Users can get rather similar things from [`questionr`](https://github.com/juba/questionr), by @juba, and [`labelled`](https://github.com/larmarange/labelled), via @larmarange, but users do not necessarily know those packages. Another similar function, although broader in scope as it search the entire environment, is [`lookfor`](https://github.com/leeper/lookfor) by @leeper.

Minimal search support through both kinds of variable labels in `haven` would be, to my view, hugely beneficial to all those users that I see using `haven` and then returning to either SPSS or Stata just to explore the data, especially when there are lots of variables with uninformative names, which happens often in large surveys like those documented by @ajdamico in [`asdfree`](https://github.com/ajdamico/asdfree).

* * *

Please let me know if adding that kind of interactivity to `haven` would be welcome, and I'll submit a PR implementing what I described above.

hadley:
This is unfortunately out of scope for haven — I have relatively bandwidth for this package so I want to keep it as simple as possible.

Issue #443, opened 2019-04-13
Add print_variable_labels convenience function


briatte:
Hello,

Would you consider a complement to `print_labels`, which is described as follows in the documentation?

> This is a convenience function, useful to explore the variables of a newly imported dataset.

Since many labelled datasets have very large numbers of variables that the user might want to list and search, here's a function that builds on `print_labels` and on the code posted by @hadley in #237 (except missing variables are returned as `NA` instead of `""`:

```r
#' Print the variable labels of a data frame
#' 
#' This is a convenience function, useful to explore the variables of a newly 
#' imported dataset.
#' @param x A data frame
#' @param name The name of the data frame (optional)
print_variable_labels <- function (x, name = NULL) {
  if (!is.data.frame(x)) {
    stop("x must be a data frame", call. = FALSE)
  }
  var_label <- function(x) {
    label <- attr(x, "label")
    if (is.null(label)) NA_character_ else label
  }
  labels <- vapply(x, var_label, character(1))
  x <- names(x)
  names(x) <- labels
  cat("\nVariable labels:", name, "\n", sep = "")
  lab_df <- data.frame(value = x, label = labels, stringsAsFactors = FALSE)
  print(lab_df, row.names = FALSE)
  invisible(x)
}
```

Example with `iris.dta`

```r
ex <- read_dta(system.file("examples", "iris.dta", package = "haven"))
print_variable_labels(ex)
```

```
> print_variable_labels(ex)

Variable labels:
       value        label
 sepallength Sepal.Length
  sepalwidth  Sepal.Width
 petallength Petal.Length
  petalwidth  Petal.Width
     species      Species
```

The output can get unwieldy if there are many variables, or if the labels are long.

The (invisible) result might be stored for further exploration:

```r
lbl <- print_variable_labels(ex)
lbl[ grepl("Petal", names(lbl)) ]
```

```
> lbl[ grepl("Petal", names(lbl)) ]
 Petal.Length   Petal.Width 
"petallength"  "petalwidth" 
```

This function is only a small step to helping users with interactive exploration labelled data (such as survey data), but perhaps a minimal implementation is good enough for `haven`, leaving it to other packages like `questionr` or `labelled` to provide more.

I will of course submit a PR if that function is of interest.

hadley:
Again, I unfortunately think this is out of scope for haven. It really is a package designed for reading and writing statistical file formats, not generally computing on the data representations used by those packages.

Issue #243, opened 2016-11-28
Others have been able to create SAS datasets from R but still fails for me


rogerjdeangelis:
Does any one know the combination of operating systems, bitness.. where SAS/WPS/Stat Transfer or Carolina can read a R created SAS dataset. This is a different test that a round trip in R.

Here is my latest try

  I have been unable to create a SAS dataset using Win 7 64bit, SAS 64bit 9.4M2 and R 64bit 3.2.4. It appears that others have been successful.

I have also tried with the latest version of R.

I get the error that 'it is not a SAS dataset' and I cannot do a 'proc contents' , repair or datastep 

see

https://goo.gl/549bbM
http://www.oralytics.com/2016/11/creating-and-reading-spss-and-sas-data.html#comment-form

Here is one of my attempts

> library(haven)
> write_sas(mtcars, "d:/sd1/cars.sas7bdat")
>
NOTE: 2 lines were written to file PRINT.
Stderr output:
Warning message:
package 'haven' was built under R version 3.2.5

proc contents data=sd1.cars;
run;quit;

1930  proc contents data=sd1.cars;
ERROR: File SD1.CARS.DATA is not a SAS data set.
1931  run;

data tst;
  set sd1.cars;
run;quit;

932  data tst;
933    set sd1.cars;
RROR: File SD1.CARS.DATA is not a SAS data set.
934  run;

OTE: The SAS System stopped processing this step because of errors.
ARNING: The data set WORK.TST may be incomplete.
When this step was stopped there were 0 observations and 0 variables.
ARNING: Data set WORK.TST was not replaced because this step was stopped.
OTE: DATA statement used (Total process time):

Other attempys
proc datasets lib=sd1;
repair cars;
run;quit;

Libref         SD1
Engine         V9
Physical Name  d:\sd1
Filename       d:\sd1
ERROR: File SD1.CARS.DATA is not a SAS data set.
ERROR: File SD1.CLASSR.DATA is not a SAS data set.
ERROR: File SD1.MTCARS.DATA is not a SAS data set.
ERROR: File SD1.TESTFIX.DATA is not a SAS data set.
ERROR: File SD1.TESTR.DATA is not a SAS data set.
ERROR: File SD1.TESTRR.DATA is not a SAS data set.
ERROR: File SD1.X.DATA is not a SAS data set.

PR #446, opened 2019-04-15
Update labelled.R with search functions

briatte:
Dear package maintainers,

As suggested in #243, #443 and #444, this PR adds search functions to let the user print and find value and variable labels as easily as possible.

Target audience: typically, survey data users faced with a labelled dataset imported from SPSS or Stata, with 100s of uninformatively named variables as in `v1, v2, … v100` or `Q3a`, `q77` etc. Try importing virtually any survey from an official statistical agency to get an example.

## Example usage

Using a labelled dataset from the [`questionr`](https://github.com/juba/questionr) package:

```r
# library(haven)
> data("fertility", package = "questionr")
> ex <- women
```
Let's start with __variable labels__:
```r
print_variable_labels(ex)
```
```
Variable labels: 
# A tibble: 17 x 2
   variable          label                             
   <chr>             <chr>                             
 1 id_woman          Woman Id                          
 2 id_household      Household Id                      
 3 weight            Sample weight                     
 4 interview_date    Interview date                    
 5 date_of_birth     Date of birth                     
 6 age               Age at last anniversary (in years)
 7 residency         Urban / rural residency           
 8 region            Region                            
 9 instruction       Level of instruction              
10 employed          Employed?                         
11 matri             Matrimonial status                
12 religion          Religion                          
13 newspaper         Read newspaper?                   
14 radio             Listen to radio?                  
15 tv                Watch TV?                         
16 ideal_nb_children Ideal number of children          
17 test              Ever tested for HIV?              
```
Let's grep a few variable labels (the function currently only greps those, not names):

```r
find_variable_labels(women, "wom.n|child|hiv")
```
```
# A tibble: 3 x 2
  variable          label                   
  <chr>             <chr>                   
1 id_woman          Woman Id                
2 ideal_nb_children Ideal number of children
3 test              Ever tested for HIV?    
```
If nothing's found:
```r
find_variable_labels(women, "xyz")
```
```
# A tibble: 0 x 2
# … with 2 variables: variable <chr>, label <chr>
```
Onto __value labels__:
```r
print_labels(ex$instruction)
```
```
Labels: 
# A tibble: 4 x 2
  value label    
  <chr> <chr>    
1 0     none     
2 1     primary  
3 2     secondary
4 3     higher   
```
Silly grep example:
```r
find_labels(women$instruction, "RY")
```
```
# A tibble: 2 x 2
  value label    
  <chr> <chr>    
1 1     primary  
2 2     secondary
```
If nothing's found:
```r
find_labels(women$instruction, "xyz")
```
```
# A tibble: 0 x 2
# … with 2 variables: value <chr>, label <chr>
```

The examples in the PR use either made-up examples or use the `iris.dta` dataset provided with `haven`, but those are less illustrative than the example above, which comes closer to a real use-case.

## Breaking changes

`print_labels` invisibly returns the variable labels instead of the actual variable, as `haven` currently does. I understand this will break continuous integration, so can revert to current behaviour if needed.

## Improvements for review

- I can write tests if need be.
- I can try to use `tidyselect` as suggested in #444, although it is not yet a dependency of `haven`.
- Anything you like :)

Best wishes,

François

### DIFF
--- a/R/labelled.R
+++ b/R/labelled.R
@@ -100,6 +100,23 @@ print.haven_labelled <- function(x, ..., digits = getOption("digits")) {
   invisible()
 }
 
+# -- value labels --------------------------------------------------------------
+
+#' @importFrom tibble tibble
+enframe_labels <- function(x) {
+  if (!is.labelled(x)) {
+    stop("x must be a labelled vector", call. = FALSE)
+  }
+  labels <- attr(x, "labels", exact = TRUE)
+  if (length(labels) == 0) {
+    return(invisible(x))
+  }
+  value <- if (is.double(labels))
+    format_tagged_na(labels)
+  else unname(labels)
+  tibble::tibble(value = value, label = names(labels))
+}
+
 #' Print the labels of a labelled vector
 #'
 #' This is a convenience function, useful to explore the variables of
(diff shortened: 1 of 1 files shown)


hadley:
Per the issue discussion, this isn't something that we're currently looking for in haven, but these would make a great contribution to another package!

PR #459, opened 2019-06-24
fix argument name in labelled help, closes #458

### LINKED ISSUE #458: "labels = " should be "levels =" in labelled help
The following code snippet is taken from the help for `haven::labelled`. It used to run fine. after the update to ellipsis, it causes an error.. However, this appears to just be a bug in haven that wasn't caught before.

To fix it, we should change `labels` to `levels` in that snippet.

``` r
s1 <- haven::labelled(c("M", "M", "F"), c(Male = "M", Female = "F"))
forcats::as_factor(s1, labels = "values")
#> 1 components of `...` were not used.
#> 
#> We detected these problematic arguments:
#> * `labels`
#> 
#> Did you misspecify an argument?
```

<sup>Created on 2019-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

rubenarslan:


hadley:
Thanks!

PR #460, opened 2019-07-02
First attempt at eliminating strncpy warnings

hadley:


src/readstat/sas/readstat_sas.c
```diff
@@ -281,7 +281,8 @@ readstat_error_t sas_write_header(readstat_writer_t *writer, sas_header_info_t *
         .host = "W32_VSPRO"
     };
 
-    strncpy(header_start.file_label, writer->file_label, sizeof(header_start.file_label));
+    memset(header_start.file_label, ' ', sizeof(header_start.file_label));
+    memcpy(header_start.file_label, writer->file_label, sizeof(header_start.file_label));
```
hadley: This doesn't seem right since I fill with blanks and then immediately overwrite with the same number of bytes.

Also `writer->file_label` is `char file_label[100]` and `header_stat.file_label` is `char file_label[64]`

src/readstat/sas/readstat_sas.c
```diff
@@ -334,7 +335,8 @@ readstat_error_t sas_write_header(readstat_writer_t *writer, sas_header_info_t *
 
     char release[32];
     snprintf(release, sizeof(release), "%1ld.%04dM0", writer->version, 101);
-    strncpy(header_end.release, release, sizeof(header_end.release));
+    memset(header_end.release, ' ', sizeof(header_end.release));
```
hadley: I think I should be zero-filling `release`, not `header_end.release`

src/readstat/spss/readstat_sav_write.c
```diff
@@ -141,28 +141,30 @@ static readstat_error_t sav_emit_header(readstat_writer_t *writer) {
...
 
     char creation_date[sizeof(header.creation_date)+1];
     snprintf(creation_date, sizeof(creation_date),
-            "%02d %3.3s %02d", 
+            "%02d %3.3s %02d",
             (unsigned int)time_s->tm_mday % 100,
             months[time_s->tm_mon],
             (unsigned int)time_s->tm_year % 100);
-    strncpy(header.creation_date, creation_date, sizeof(header.creation_date));
+    memset(header.creation_date, ' ', sizeof(header.creation_date));
```
hadley: Similarly here and below

hadley:
Need to fix whitespace changes as described in https://stackoverflow.com/questions/13793241/

src/readstat/readstat_writer.c
```diff
@@ -78,15 +78,15 @@ static void readstat_label_set_free(readstat_label_set_t *label_set) {
 static void readstat_copy_label(readstat_value_label_t *value_label, const char *label) {
     if (label && strlen(label)) {
         value_label->label_len = strlen(label);
-        value_label->label = malloc(value_label->label_len);
-        strncpy(value_label->label, label, value_label->label_len);
+        value_label->label = calloc(1, value_label->label_len);
+        memcpy(value_label->label, label, value_label->label_len);
     }
```
gaborcsardi: Is this just `strdup()`?

gaborcsardi: Maybe you don't want to write the trailing zero? In that case it is not `strdup()`, and the `memcpy()` is fine.

hadley: Yeah, this is creating a pascal-style string (i.e. array of data + length)

src/readstat/readstat_writer.c
```diff
@@ -265,15 +265,15 @@ readstat_error_t readstat_write_space_padded_string(readstat_writer_t *writer, c
...
                 writer->label_sets_capacity * sizeof(readstat_label_set_t *));
     }
     readstat_label_set_t *new_label_set = calloc(1, sizeof(readstat_label_set_t));
-    
+
     writer->label_sets[writer->label_sets_count++] = new_label_set;
 
     new_label_set->type = type;
-    strncpy(new_label_set->name, name, sizeof(new_label_set->name));
+    memcpy(new_label_set->name, name, sizeof(new_label_set->name));
```
gaborcsardi: This is not ideal, because it is not clear that `name` have at least `sizeof(new_label_set->name)` bytes. If it does not, that's an invalid read.

src/readstat/readstat_writer.c
gaborcsardi: Btw `malloc()` or `calloc()` does not matter here, because you overwrite the whole region. So no need to change the `malloc()` line.

src/readstat/readstat_writer.c
```diff
@@ -327,8 +327,8 @@ void readstat_label_string_value(readstat_label_set_t *label_set, const char *va
     readstat_value_label_t *new_value_label = readstat_add_value_label(label_set, label);
     if (value && strlen(value)) {
         new_value_label->string_key_len = strlen(value);
-        new_value_label->string_key = malloc(new_value_label->string_key_len);
-        strncpy(new_value_label->string_key, value, new_value_label->string_key_len);
+        new_value_label->string_key = calloc(1, new_value_label->string_key_len);
+        memcpy(new_value_label->string_key, value, new_value_label->string_key_len);
     }
```
gaborcsardi: This is fine, but again, the `malloc()` change is not needed.

src/readstat/sas/readstat_sas.c
gaborcsardi: Yeah, I think memset is not needed, and I would just add a comment that `writer->file_label` is longer.

src/readstat/sas/readstat_sas.c
gaborcsardi: but you overwrite the whole thing, so no filling is needed, no?

src/readstat/sas/readstat_sas.c
```diff
@@ -334,7 +335,8 @@ readstat_error_t sas_write_header(readstat_writer_t *writer, sas_header_info_t *
...
+    memcpy(header_end.release, release, sizeof(header_end.release));
```
gaborcsardi: I think this is fine, possibly with a comment the `header_end.release` is shorter.

src/readstat/spss/readstat_por_write.c
```diff
@@ -202,8 +202,8 @@ static readstat_error_t por_emit_header(readstat_writer_t *writer, por_write_ctx
     char vanity[5][40];
     memset(vanity, '0', sizeof(vanity));
 
-    strncpy(vanity[1], "ASCII SPSS PORT FILE", 20);
-    strncpy(vanity[1] + 20, writer->file_label, 20);
+    memcpy(vanity[1], "ASCII SPSS PORT FILE", 20);
+    memcpy(vanity[1] + 20, writer->file_label, 20);
     if (file_label_len < 20)
```
gaborcsardi: For clarity, you could have the string literal as a constant, and then use `sizeof()` instead of 20, but this is minor.

src/readstat/spss/readstat_sav_write.c
gaborcsardi: yeah, no `memset()` is needed.

src/readstat/spss/readstat_sav_write.c
```diff
@@ -141,28 +141,30 @@ static readstat_error_t sav_emit_header(readstat_writer_t *writer) {
...
     char creation_time[sizeof(header.creation_time)+1];
     snprintf(creation_time, sizeof(creation_time),
             "%02d:%02d:%02d",
             (unsigned int)time_s->tm_hour % 100,
             (unsigned int)time_s->tm_min % 100,
             (unsigned int)time_s->tm_sec % 100);
-    strncpy(header.creation_time, creation_time, sizeof(header.creation_time));
-    
+    memset(header.creation_time, ' ', sizeof(header.creation_time));
+    memcpy(header.creation_time, creation_time, sizeof(header.creation_time));
```
gaborcsardi: same here.

src/readstat/spss/readstat_por_write.c
gaborcsardi: But then you need `sizeof() - 1`, so maybe that just complicates it.

src/readstat/readstat_writer.c
```diff
@@ -265,15 +265,15 @@ readstat_error_t readstat_write_space_padded_string(readstat_writer_t *writer, c
...
                 writer->label_sets_capacity * sizeof(readstat_label_set_t *));
     }
     readstat_label_set_t *new_label_set = calloc(1, sizeof(readstat_label_set_t));
-    
+
     writer->label_sets[writer->label_sets_count++] = new_label_set;
 
     new_label_set->type = type;
-    strncpy(new_label_set->name, name, sizeof(new_label_set->name));
+    memcpy(new_label_set->name, name, sizeof(new_label_set->name));
```
jeroen: I think this will segfault when `strlen(name)` < `sizeof(new_label_set->name)` ?  Because the old `strncpy()` implementation would stop reading at the end of the src string `name` and then continue copying zero's.

However memcpy will keep reading past the end of the input string until `sizeof(new_label_set->name)` bytes.

So unless `const char *name` is always exactly the same size as `new_label_set->name` (which I cannot immediately see), you should not copy more bytes then `strlen(name)`.

jeroen:
Whenever you use `memcpy(dst, src, len)` you have to make sure that `len` is smaller than or equal to the size of both `dst` and `src` otherwise you will either read or write outside of the buffer. 

Perhaps you're already taking this into account but it's not immediately apparent at all the places where you are using `memcpy` instead of `strncpy` that the input is guaranteed to be large enough.

src/readstat/readstat_writer.c
gaborcsardi: OTOH there was a compiler warning for this:
```
readstat/readstat_writer.c:276:5: warning: ‘strncpy’ specified bound 256 equals destination size [-Wstringop-truncation]
     strncpy(new_label_set->name, name, sizeof(new_label_set->name));
```
so it must be always equal? If this is the case, it is worth commenting why.

src/readstat/sas/readstat_sas.c
hadley: I don't think `snprintf` is gauranteed to fill `release`?

src/readstat/readstat_writer.c
hadley: Ah yeah, of course.

src/readstat/readstat_writer.c
hadley: So something like 

```c
memcpy(new_label_set->name, name, min(strlen(name), sizeof(new_label_set->name)))
```

?

hadley:
Yeah, it makes me a little worried that carefully analysis is needed to tell whether or not this code is safe. I'll see what the upstream author has to say, when I push the patch back up.

Issue #465, opened 2019-07-28
Seems missing `-liconv` in linux build configure


lguangyu:
HOW...
```
# wget https://cloud.r-project.org/src/contrib/haven_2.1.1.tar.gz
# untar
R CMD INSTALL ./haven
```

WHAT...
Failed load test:
```
** testing if installed package can be loaded
Error: package or namespace load failed for ‘haven’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/home/li.gua/usr/R/haven/libs/haven.so':
  /home/li.gua/usr/R/haven/libs/haven.so: undefined symbol: libiconv
Error: loading failed
Execution halted
ERROR: loading failed
```
`ldd` check confirms haven.so not linked into `libiconv.so`:
```bash
$ ldd src/haven.so 
	linux-vdso.so.1 =>  (0x00007ffd29276000)
	libz.so.1 => /home/li.gua/.local/lib/libz.so.1 (0x00002b4226b7f000)
	libR.so => not found
	libstdc++.so.6 => /shared/centos7/gcc/7.2.0/lib64/libstdc++.so.6 (0x00002b4226d9c000)
	libm.so.6 => /lib64/libm.so.6 (0x00002b422711d000)
	libgcc_s.so.1 => /shared/centos7/gcc/7.2.0/lib64/libgcc_s.so.1 (0x00002b422741f000)
	libc.so.6 => /lib64/libc.so.6 (0x00002b4227636000)
	/lib64/ld-linux-x86-64.so.2 (0x00002b42266fa000)
```

`src/Makevars`:
```
CFILES = $(wildcard *.c readstat/*.c readstat/sas/*.c readstat/spss/*.c readstat/stata/*.c)
CPPFILES = $(wildcard *.cpp)

SOURCES = $(CFILES) $(CPPFILES)

# This must be defined identically in Makevars.win
OBJECTS = $(CFILES:.c=.o) $(CPPFILES:.cpp=.o)

PKG_CFLAGS = -Ireadstat -DHAVE_ZLIB
PKG_CXXFLAGS = -Ireadstat -DHAVE_ZLIB
PKG_LIBS = -lz
```

Manually add `-liconv` to `PKG_LIBS` temporarily solve the problem.

hadley:
It builds fine for us, so this must be something peculiar to your system.

liutiming:
[...]

PR #470, opened 2019-09-18
Add -liconv to src/Makevars

pvanheus:
This addresses the problem reported in https://github.com/tidyverse/haven/issues/465 which causes install problems on Linux.


hadley:
I'm suspicious this is the correct fix because, as far as I can tell, `-liconv` only appears in `Makevars.win` of CRAN packages: https://github.com/search?p=2&q=org%3Acran+liconv+filename%3AMakevars&type=Code

Issue #475, opened 2019-10-17
BUG: tibbles cannot be printed when labels are duplicated


courtiol:
When labels are duplicated in a column of class "haven_labelled", tibbles can no longer be displayed because of an inner call to `as_factor(., "label")` which fails:

``` r
library(haven)
foo <- labelled(1:3, labels = c("A" = 1, "B" = 2, "A" = 3))
as_factor(foo, "default") # works
#> [1] A B A
#> Levels: A B
as_factor(foo, "label") # fails
#> Error in `levels<-`(`*tmp*`, value = as.character(levels)): factor level [3] is duplicated

library(tibble)
bar <- tibble(x = foo)
bar
#> Error in `levels<-`(`*tmp*`, value = as.character(levels)): factor level [3] is duplicated
```
The culprit is the function `as_factor.haven_labelled`.
I will PR what I think is a patch that solves this issue.

I am not an SPSS expert but I discussed the problem with a few SPSS users and it seems that using repetition in the labels is something that they do.

courtiol:
Duplicate #424

PR #476, opened 2019-10-17
Add support for duplicated levels in as_factor

courtiol:
This addresses the issue #475 
The solution involves only adding a call to `unique()` around the labels when calling `factor()`.
I created a test that shows the problem in the original implementation.
I updated the NEWS file.


courtiol:
Hummm... I realized too late this is a duplicate from #425 
Since we came up with virtually the same solution, I close it.

PR #524, opened 2020-05-29
Use cpp11

jimhester:


jimhester:
This ended up being pretty straightforward, I opened a PR here as this gives a good example to discuss if we want to change class names, add additional API points etc.

src/DfReader.cpp
```diff
@@ -155,15 +167,15 @@ class DfReader {
...
+    output_.resize(ncols_);
+    names_.resize(ncols_);
     val_labels_.resize(ncols_);
     var_types_.resize(ncols_);
   }
 
   void setMetadata(const char *file_label) {
     if (file_label != NULL && strcmp(file_label, "") != 0) {
-      output_.attr("label") = CharacterVector::create(Rf_mkCharCE(file_label, CE_UTF8));
+      output_.attr("label") = file_label;
```
hadley, line 178: Nice!

src/DfReader.cpp
```diff
@@ -188,22 +200,22 @@ class DfReader {
     switch(readstat_variable_get_type(variable)) {
     case READSTAT_TYPE_STRING_REF:
     case READSTAT_TYPE_STRING:
-      output_[var_index] = CharacterVector(nrowsAlloc_);
+      output_[var_index] = tidycpp::safe[Rf_allocVector](STRSXP, nrowsAlloc_);
```
hadley, line 203: Why can't you use the tidycpp equivalent of `CharacterVector` here?

src/DfReader.cpp
```diff
@@ -237,7 +249,8 @@ class DfReader {
       case READSTAT_TYPE_STRING_REF:
       case READSTAT_TYPE_STRING:
       {
-        CharacterVector na_values(n_ranges);
+        tidycpp::writable::character_vector na_values;
```
hadley, line 252: Should this have a constructor that takes the length?

src/DfReader.cpp
```diff
@@ -246,7 +259,7 @@ class DfReader {
         }
 
         col.attr("na_values") = na_values;
-        col.attr("class") = CharacterVector::create("haven_labelled_spss", "haven_labelled", "vctrs_vctr", "character");
+        col.attr("class") = tidycpp::as_sexp({"haven_labelled_spss", "haven_labelled", "vctrs_vctr", "character"});
```
hadley, line 262: Why does this need an explicit `as_sexp()`?

src/DfReader.cpp
```diff
@@ -280,19 +294,19 @@ class DfReader {
...
+        col.attr("class") = {"haven_labelled_spss", "haven_labelled", "vctrs_vctr", "double"};
       }
       }
     }
 
 
     // Store original format as attribute
     if (var_format != NULL && strcmp(var_format, "") != 0) {
-      col.attr(formatAttribute(vendor_)) = Rf_ScalarString(Rf_mkCharCE(var_format, CE_UTF8));
+      col.attr(formatAttribute(vendor_).c_str()) = var_format;
```
hadley, line 309: Should `.attr()` have a `std::string()` variant?

src/DfWriter.cpp
```diff
@@ -175,18 +178,18 @@ class Writer {
...
       return NULL;
 
     return string_utf8(label, 0);
   }
 
-  const char* var_format(RObject x, VarType varType) {
+  const char* var_format(tidycpp::sexp x, VarType varType) {
     // Use attribute, if present
-    RObject format = x.attr(formatAttribute(vendor_));
+    tidycpp::sexp format(x.attr(formatAttribute(vendor_).c_str()));
```
hadley, line 192: Can you please remind me why we can't use `=` here?

src/DfReader.cpp
```diff
@@ -451,7 +465,7 @@ int dfreader_value_label(const char *val_labels, readstat_value_t value,
 }
 
 void print_error(const char* error_message, void* ctx) {
-  Rcout << error_message << "\n";
+  Rprintf("%s\n", error_message);
```
hadley, line 468: Worth having `tidycpp::cat()` to parallel `tidy::stop()`?

src/DfReader.cpp
jimhester: I can now, I just switched it. When I initially wrote that part I didn't have a constructor that took a size. My thought was that `std::vector()` didn't have an equivalent constructor so I was following that API. 

But after doing this conversion think it does make sense to match Rcpp have one, since it is pretty common to have to pre-allocate the vector sizes, so I have added it.

src/DfReader.cpp
jimhester: It actually doesn't, I have removed it.

src/DfWriter.cpp
jimhester: sexp didn't have a default constructor initially, `foo x = y` generally default constructs x and then using the assignment operator, `foo x(y)` uses the constructor that takes the `y` object.

I ended up having to add a default constructor for `sexp` objects to handle cases like https://github.com/jimhester/haven/blob/36d6c6e9e200469f4b3cb5aaaf9e12848278dfcb/src/DfReader.cpp#L84., so it could potentially change now.

However it is a bit more efficient to call the constructor directly.

src/DfReader.cpp
jimhester: Maybe it should...

jimhester:
This is now updated to use cpp11

jimhester:
There is a test failure only on R 3.4 and 3.3 (https://github.com/tidyverse/haven/runs/766643625#step:10:135)

I am not really clear if this is related to this PR or due to some difference in the tibble stack for those versions.

hadley:
It's weird that the messages seem to be emitted, but they're not caught by the test.

jimhester:
Oh, possibly because we have to use TopLevelExec in R < 3.5 when there is no UnwindProtect available

jimhester:
Ok, the tests are now passing for this, I think it can be merged whenever.

PR #544, opened 2020-09-06
Use old-style SPSS compression by default for .sav (SPSS) files, sinc…

oliverbock:
…e almost nobody uses the inefficient uncompressed format.  The newer ZCompressed format (compress=TRUE) is even better, but is much less used and less supported by third-party tools.

### DIFF
--- a/man/read_spss.Rd
+++ b/man/read_spss.Rd
@@ -91,7 +91,7 @@ to enforce them.}
 \item{path}{Path to a file where the data will be written.}
 
 \item{compress}{If \code{TRUE}, will compress the file, resulting in a \code{.zsav}
-file.}
+file.  Otherwise the \code{.sav} file will be bytecode compressed.}
 }
 \value{
 A tibble, data frame variant with nice defaults.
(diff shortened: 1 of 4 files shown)


hadley:
Could you please update the documentation for `write_sav()` and add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

man/read_spss.Rd
```diff
@@ -91,7 +91,7 @@ to enforce them.}
 \item{path}{Path to a file where the data will be written.}
 
 \item{compress}{If \code{TRUE}, will compress the file, resulting in a \code{.zsav}
-file.}
+file.  Otherwise the \code{.sav} file will be bytecode compressed.}
```
hadley, line 94: You need to edit the underlying roxygen2 docs next to the definition of `read_spss()`

oliverbock: Ah, got it.  Thanks.

hadley:
Thanks!

Issue #576, opened 2021-04-07
read_sav returns `Error: Invalid input type, expected 'list' actual 'NULL'`


sjkiss:
I tried installing the development version of `haven` to address [this](https://github.com/tidyverse/dplyr/issues/5196) problem, but `read_sav` seems to no longer read in this data file properly. It returns `Error: Invalid input type, expected 'list' actual 'NULL'`

```r
remotes::install_github('tidyverse/haven')
library(haven
dat<-read_sav('https://github.com/LISPOP/public-health-canada/raw/main/data/ORD-571352-N3W7_Final_SPSS_032221.SAV')

```

sjkiss:
Can confirm: jumping back to haven 2.3.1 reads the file in properly.

hadley:
Underlying bug in cpp11: https://github.com/r-lib/cpp11/issues/166

Issue #557, opened 2020-12-04
Infinite recursion on bind_rows for labelled_spss vectors


gorcha:
The `haven_labelled_spss` class ends up in an infinite loop when using `bind_rows()`.
I think this is because the vctrs casting/coercion methods haven't been implemented for `haven_labelled_spss` as they have for `haven_labelled`.

``` r
library(dplyr, warn.conflicts = FALSE)
library(haven)

x <- tibble(a = labelled(1))
x
#> # A tibble: 1 x 1
#>           a
#>   <dbl+lbl>
#> 1         1
bind_rows(x, x)
#> # A tibble: 2 x 1
#>           a
#>   <dbl+lbl>
#> 1         1
#> 2         1
rbind(x, x)
#> # A tibble: 2 x 1
#>           a
#>   <dbl+lbl>
#> 1         1
#> 2         1

x <- tibble(a = labelled_spss(1))
x
#> # A tibble: 1 x 1
#>           a
#>   <dbl+lbl>
#> 1         1
bind_rows(x, x)
#> Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
rbind(x, x)
#> # A tibble: 2 x 1
#>           a
#>   <dbl+lbl>
#> 1         1
#> 2         1
```

<sup>Created on 2020-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Issue #538, opened 2020-07-29
read_spss (user_na = TRUE) leads to: Can't combine `..1` <labelled_spss<double>> and `..2` <double> 


bhaney22:
Setting `user_na = TRUE`  in the haven::read_spss() command creates a file that produces an error when used subsequently in a dplyr::mutate() command.  

The following reprex has two examples: the first example, mygss1, shows that haven::read_spss() with user_na=FALSE (the default) works as expected. It creates a file that can be manipulated with dplyr::mutate() without error.

The second example, mygss2, is created by using haven::read_spss(..., user_na = TRUE) and leads to an error when that file is subsequently used by dplyr::mutate.

This issue was originally submitted as an issue for dplyr (https://github.com/tidyverse/dplyr/issues/5424). That issue was closed because it is appeared to be an issue with a different package and not with dplyr. Thus, this issue is being raised for the haven package. I notice that:

The `user_na = FALSE` option leads to the list of classes for the variables in the dataset to be:

` $ class      : chr [1:3] "haven_labelled" "vctrs_vctr" "double"`

While, the `user_na = TRUE` option leads to the list of classes for the variables in the dataset to be:

` $ class      : chr [1:4] "haven_labelled_spss" "haven_labelled" "vctrs_vctr" "double"`

Since UseMethod() only considers the first two classes of the object for dispatch, could the additional class in the front of the list of classes be causing the error? 

``` r

library(dplyr)     
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(haven) 
  
# Download the 2018 GSS dataset in SPSS format from the GSS website:

temp1 <- tempfile()
download.file("https://gss.norc.org/Documents/spss/2018_spss.zip", temp1)
unzip(temp1,"GSS2018.sav")
unlink(temp1)

# Read in SPSS using haven without user_na option (this works as expected)

mygss1 <- haven::read_spss("GSS2018.sav", col_select = c("RACE","TAX"))

summary(mygss1)
#>       RACE            TAX       
#>  Min.   :1.000   Min.   :1.000  
#>  1st Qu.:1.000   1st Qu.:1.000  
#>  Median :1.000   Median :1.000  
#>  Mean   :1.394   Mean   :1.483  
#>  3rd Qu.:2.000   3rd Qu.:2.000  
[...]

Issue #534, opened 2020-07-14
labelled_spss: Can't convert <labelled_spss<double>> to <double>


zahlenzauber:
Thanks for the haven-package! There is a small issue: With labelled_spss instead of labelled I get an error messages for some (but not all) functions when I use the variable.
I guess it has something to do with the class of the object or maybe I just got something wrong.
Please see the MWE below.

I use R version 4.0.1 (2020-06-06)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363)

```r


rm(list=ls())

set.seed(5)


data <- data.frame(var1=sample(1:5,50,replace=T) , var2=sample(1:5,50,replace=T))

# as expected
data$var1 <- haven::labelled(data$var1, labels=c("very low"=1, "low"=2, "medium"=3, "high"=4, "very high"=5))
sd(data$var1)
mean(data$var1)

# for mean as expected, for sd error "Can't convert <labelled_spss<double>> to <double>."
data$var2 <- haven::labelled_spss(data$var2, labels=c("very low"=1, "low"=2, "medium"=3, "high"=4, "very high"=5))
mean(data$var2)
sd(data$var2)
```

atungate:
I am able to reproduce the error:

R version 4.0.0 (2020-04-24)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363)

haven_2.3.1  


```
vec <-  haven::labelled_spss(sample(1:5,50,replace=T), labels=c("very low"=1, "low"=2, "medium"=3, "high"=4, "very high"=5))
> as.double(vec) 
Error: Can't convert <labelled_spss<integer>> to <double>.
Run `rlang::last_error()` to see where the error occurred.

> rlang::last_error()
<error/vctrs_error_incompatible_type>
Can't convert <labelled_spss<integer>> to <double>.
Backtrace:
 1. base::as.double(vec)
 4. vctrs::vec_default_cast(...)
 5. vctrs::stop_incompatible_cast(...)
 6. vctrs::stop_incompatible_type(...)
 7. vctrs:::stop_incompatible(...)
 8. vctrs:::stop_vctrs(...)
Run `rlang::last_trace()` to see the full context.

> rlang::last_trace()
<error/vctrs_error_incompatible_type>
Can't convert <labelled_spss<integer>> to <double>.
Backtrace:
    x
 1. +-base::as.double(vec)
[...]

PR #577, opened 2021-04-08
Add vctrs support for haven_labelled_spss

gorcha:
@hadley, I've implemented the vctrs methods for haven_labelled_spss.
This should fix at least a few issues - #527, #534, #538, and #557.

I've mostly stolen from the original work done on haven_labelled, but there are a few things to note:
* If casts from one labelled_spss change the user missing values/range they're only treated as lossy if data that were missing are no longer missing. I wasn't 100% sure how to treat this so went with this approach because it's in line with the treatment of value labels.
* Coercion generally follows the same rules as labelled vectors
* Combining two labelled_spss vectors with different labels will do the same thing as labelled (i.e. revert to the base class)
* Combining two labelled_spss vectors with the same labels but non-identical missing values/range attributes will drop these attributes and return a labelled vector

It looks like the C++ bits aren't working at the moment (maybe related to changes in cpp11?) so some tests are failing at the moment but they're unrelated to the changes (also see #576).


hadley:
@gorcha I should've fixed the build failures. 

Honestly, I have very little recollection of how this code works, and it's probably not worth my time right now to fully regain that knowledge, given that this is likely to be a major improvement even if it's not 100% correct. So is there anything in particularly that you think I should look at?

gorcha:
Thanks @hadley!

Nothing in particular that needs attention, mostly just after validation that I haven't done anything dumb :slightly_smiling_face:, particularly since it's a bit of an oddball class.

hadley:
Ok, lets just merge it then. It already resolves the biggest issues folks are having with haven, and we can resolve the smaller issues as they come in.

Issue #606, opened 2021-05-20
`write_dta` can't save numeric labelled variables without value labels


domingoscisci:
Hi all,
I have a data.frame of double (no integer) variables that I want to label (only variable label, not value labels). If I try to use the `labelled()` function, a vector of class _dbl+lbl_ is correctly created, but it cannot be saved in a Stata file using `write_dta()`. If I use the `attr()` function the problem does not arise.
It is probably not correct to use `labelled()` to label the variable only. In this case, perhaps it might be useful to make this explicit within the documentation (maybe I missed something?).

Here is a reprex:

```r
library(tibble)
library(haven)


dat <- tibble(
  a = c(0.1, 0.2, 0.3, 0.4, 0.5)
)

# label variable with haven::labelled
dat$a <- labelled(dat$a, label = "test label")

# write_dta fails
write_dta(dat, tempfile(fileext = "dta"))
#> Error: Stata only supports labelling with integers.
#> Problems: `a`


dat <- tibble(
  a = c(0.1, 0.2, 0.3, 0.4, 0.5)
)

# label variable with attr
attr(dat$a, "label") <- "test label"

# write_dta ok
write_dta(dat, tempfile(fileext = "dta"))
```

Thank you for your help,
Domingo

jmobrien:
Seconding this issue, and noting that it is uniquely a problem with non-integer numerics lacking value labels:

``` r
require(haven) # 2.4.1
#> Loading required package: haven
df <- 
  data.frame(
    x1 = labelled(1:10),
    x2 = labelled(as.numeric(1:10), labels = c(one = 1)),
    x3 = labelled(as.character(1:10)),
    x4 = labelled(as.numeric(1:10))
  )

path <- tempfile()
write_dta(df, path)
#> Error: Stata only supports labelling with integers.
#> Problems: `x4`
```

<sup>Created on 2021-06-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jmobrien:
Problem seems to be here:

https://github.com/tidyverse/haven/blob/dbd65a3c3aa087ecacf1a378360ad9c967954cf1/R/haven-stata.R#L150

`is_integerish()` returns FALSE if the labels are missing, producing this issue.  Submitted PR #609 to address.

dusadrian:
If you don't mind a workaround, this solves your example:

``` r
install.packages("declared")
devtools::install_github("dusadrian/DDIwR")
library(declared)
[...]

PR #609, opened 2021-06-03
Fixing write_dta labels checking, addressing #606

jmobrien:
Addresses #606.

Added an initial check for missing labels in `has_non_integer_labels()`.  Previously, `as_integerish()` was returning FALSE when labels were missing.  

Also, associated test was missing an apparently-intended line that would have caught this.  Test passes now.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # haven (development version)
 
+* `write_dta()` now correctly handles "labelled"-class numeric (double) variables 
+   that don't have value labels (@jmobrien, #606, #609).
+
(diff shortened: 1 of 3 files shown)


hadley (approved):
Thanks!  Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks @jmobrien — btw if you use "fixes" or "closes" instead of "addresses", GitHub will automatically close the associated issue when the PR is merged.

jmobrien:
Glad to help.  And thanks for the useful tip.

PR #613, opened 2021-06-22
Fix roundtrip typo

urswilke:
Instead of comparing the variables in the original dataframes to those resulting of roundtripping (by saving and reading), they were compared to themselves.


hadley:
Woops! Thanks for catching this.

Issue #608, opened 2021-05-22
read_dta failure


iamforerunner:
dta with chinese character
read_dta failure
please refer to the dta zip file attached.

R version 4.0.5
haven version 2.4.1
[a1.zip](https://github.com/tidyverse/haven/files/6526776/a1.zip)

gorcha:
Looks like the same issue as #600

Issue #600, opened 2021-05-12
RStudio crash on importing stata file


sjkiss:
There is a GitHub repo [here](https://github.com/sjkiss/ces19) that replicates the problem, but this code will do so as well.  
On reading in the Stata file, RStudio crashes immediately. I note that this happened in another [issue](https://github.com/tidyverse/haven/issues/79)

```r
library(haven)
out<-read_dta(file="https://github.com/sjkiss/ces19/raw/main/2019%20Canadian%20Election%20Study%20-%20Online%20Survey%20v1.0.dta")

```
Results from `SessionInfo()`

```
R version 4.0.4 (2021-02-15)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur 10.16

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.0/Resources/lib/libRlapack.dylib

locale:
[1] en_CA.UTF-8/en_CA.UTF-8/en_CA.UTF-8/C/en_CA.UTF-8/en_CA.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] haven_2.4.1.9000

loaded via a namespace (and not attached):
 [1] compiler_4.0.4  ellipsis_0.3.2  magrittr_2.0.1  hms_1.0.0       tools_4.0.4     pillar_1.6.0   
 [7] tibble_3.1.1    crayon_1.4.1    utf8_1.2.1      fansi_0.4.2     vctrs_0.3.8     forcats_0.5.1  
[13] lifecycle_1.0.0 pkgconfig_2.0.3 rlang_0.4.11 
```

jobreu:
Same problem here. The R session is aborted in RStudio when I try to read a .dta file using `read_dta()`. I am using R version 4.1.0 and haven version 2.4.1 with Windows 10 as OS.

gorcha:
Hi all, It looks like the bug in #79 was re-introduced following some other updates.
I've put together a simple fix, will chuck a PR up shortly.

DaniMori:
Issue #620 is a duplicate of this one.

Issue #79, opened 2015-06-23
read_dta leads to immediate R crash


jon-mellon:
When I try reading a .dta file into R using read_dta, it causes a complete crash of R. 
The following minimal file causes this crash:
https://github.com/jon-mellon/mellonMisc/blob/master/data/panel_leader.dta

Details of my system are below. Note that I am using Rstudio version 0.98.1103
               _                           
platform       x86_64-w64-mingw32          
arch           x86_64                      
os             mingw32                     
system         x86_64, mingw32             
status                                     
major          3                           
minor          2.0                         
year           2015                        
month          04                          
day            16                          
svn rev        68180                       
language       R                           
version.string R version 3.2.0 (2015-04-16)
nickname       Full of Ingredients     

> Sys.info()
       sysname        release        version       nodename        machine 
     "Windows"        "7 x64"   "build 9200"  "JON-SURFACE"       "x86-64" 
         login           user effective_user 
         "Jon"          "Jon"          "Jon"

jon-mellon:
Note that I suspect the issue is with this dataset, but it's not an issue that is causing any known problems with stata or SPSS, so I suspect many files out there may suffer from these issues too.

hadley:
I've fixed the crash, but I'm not sure I'm getting the correct value.  

What value do you see in stata for generalElectionVotePostOthW5 in row 41838?

jon-mellon:
My file only has 41837 rows (excluding headers). If you mean row 41837 (the last row in the file), then running:

list generalElectionVotePostOthW5 if id==34803

gives
```
       | g~tOthW5 |
       |----------|
41837. |          |
```

jon-mellon:
By the way, I was getting a related error when trying to read in the SPSS version of this file as well. Do you want me to open up a separate issue or will it be sorted by this one?

hadley:
Ok so that variable uses both `__NA__ ` and nothing to represent missing values? 

The SPSS issue will probably be fixed as well, but please try it out with the dev version and let me know if it doesn't work

jon-mellon:
Yes, the "__NA__" items are created by the survey company, the blanks are the result of merging versions of the datasets (where the question wasn't asked to all respondents in all waves).

Issue #594, opened 2021-04-27
R Session Abort caused by read_dta


micsky:
I was working with a r script, which I have already written several weeks ago and everything was working just fine. Today I wanted to read in a .dta file and suddendly every time I am trying to us the ````read_dta```` function to do so the session crashes. A collegue of mine is experiencing the same problem as well as another person from stackoverflow (see my question on stackoverflow - https://stackoverflow.com/questions/67288533/am-i-the-only-one-experiencing-problems-with-rs-haven-package?answertab=votes#tab-top ).

I am using haven version 2.4.1, R 4.0.5. on Windows 10 x64 (build 19042).

```r
# code which was working fine until now
install.packages("haven")
library(haven)
my_data <- read_dta("my_path_structure\\file_name.dta")

# provoked R console feedback
*** caught segfault ***
address 0x0, cause 'invalid permissions'
```

brhkim:
Hi! I just had this exact same issue, but I was finding it completely dependent on the underlying dataset I was loading in. I did some trial-and-error and found that the root cause was a couple of variables stored as type `strL`. Converting those variables to `str#`'s in Stata before saving resolved the issue while loading into R via Haven. Not sure if this will apply broadly, but that's how it worked for me.

micsky:
@brhkim Hey. Oh, that's an interesting point! Think I will try this myself.

PR #619, opened 2021-07-25
Fix issue with read_dta crashing R

gorcha:
The updates for cpp11 reintroduced the bug in #79 when read_dta is used on a file that has empty values in an StrL variable (#600, #608). This PR restores the old behaviour and returns an NA for these observations.

@hadley, the [StrL section of the dta spec](https://www.stata.com/help.cgi?dta#strls_vo) says that a NULL (i.e. no StrL address) should be interpreted as the empty string `""` rather than a missing value, so `NA_STRING` might not actually be the most appropriate value here (although it would be consistent with previous versions of haven).
What do you think?

### DIFF
--- a/src/DfReader.cpp
+++ b/src/DfReader.cpp
@@ -363,8 +363,7 @@ public:
       } else if (readstat_value_is_system_missing(value)) {
         col[obs_index] = NA_STRING;
       } else if (str_value == NULL) {
-        const char empty_string[1] = { '\0' };
-        col[obs_index] = cpp11::r_string(empty_string);
+        col[obs_index] = cpp11::r_string("");
       } else {
         col[obs_index] = cpp11::r_string(str_value);
       }
--- a/src/DfReader.cpp
+++ b/src/DfReader.cpp
@@ -363,7 +363,8 @@ public:
       } else if (readstat_value_is_system_missing(value)) {
         col[obs_index] = NA_STRING;
       } else if (str_value == NULL) {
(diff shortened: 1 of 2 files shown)


gorcha:
Woops, missed one, also covers #594

hadley:
Yeah, lets make it `""`. Thanks for looking into this!

gorcha:
No worries, done! I'm not much of a C person so it feels super hacky, but it works :stuck_out_tongue:

src/DfReader.cpp
```diff
@@ -354,15 +354,19 @@ class DfReader {
...
 
       if (readstat_value_is_tagged_missing(value)) {
         col[obs_index] = NA_STRING;
       } else if (!user_na_ && readstat_value_is_defined_missing(value, variable)) {
         col[obs_index] = NA_STRING;
       } else if (readstat_value_is_system_missing(value)) {
         col[obs_index] = NA_STRING;
+      } else if (str_value == NULL) {
+        const char empty_string[1] = { '\0' };
+        col[obs_index] = cpp11::r_string(empty_string);
```
hadley, line 367: Can't you just do `cpp11::r_string("")` ?

gorcha: Haha yep! Too much time in R treating "" and '' as equivalent :facepalm:

gorcha: Thanks!

hadley:
Thanks for working on this! Will prepare a release next week some time.

PR #622, opened 2021-07-30
Update to dev readstat

### LINKED ISSUE #620: R session crashes when trying to read (some) text values from a Stata file with `read_dta()`
I have a dataset in a `.dta` file which opens normally in Stata, and is also read normally with `read_dta()` in haven v.2.3.1.
However, when testing in versions 2.4.0 and 2.4.1, an "R Session Aborted" crash is produced.

I can't provide a "reprex", as when I try to reprex it, I obtain the message:

"Error: This reprex appears to crash R"

So here is the code to reproduce it:

```r
library(tidyverse)
library(haven)

setwd("~/../OneDrive - UAM/Workspace/R_bug_reporting_and_help")

# Both one-variable datasets are read normally
q4534 <- read_dta("data/q4534.dta")
q4554 <- read_dta("data/q4554.dta")

# If I collapse them together, write them as a .dta file and read it again,
#   it works normally
both_vars <- bind_cols(q4534, q4554)
write_dta(both_vars, path = "data/test.dta")
read_dta("data/test.dta")

# However, the dataset with the original two-variable dataset crashes:
read_dta("data/q4534_q4554.dta")
```

The three files are attached. Note that there are two variables with two cases each, and both are text values. The bug seems to happen when both are text variables; also, you may notice my working directory is a cloud storage drive (OneDrive). I tried pausing the synchronization, in case that could be causing the problem, but it happened anyway.

[data.zip](https://github.com/tidyverse/haven/files/6878503/data.zip)

hadley:
Fixes #620

@gorcha this look good to you?


gorcha:
Hey @hadley yep that definitely picks up the fix, thanks!
Have double checked and it's loading the test file in #620 successfully.

Only one minor thing - this is ReadStat 1.1.7 RC 0 rather than 1.1.6 (the StrL fix came in after the 1.1.6 release).

hadley:
Good catch, thanks!

PR #629, opened 2021-08-05
Add support for writing SPSS user missing values to integer vectors

### LINKED ISSUE #596: write_sav() fails to set missing values
Information about missing values is lost, roundtrip reprex below

```r
library(haven)
library(tibble)
test <- tibble(A = labelled_spss(1:10, c(Good = 1, Bad = 8, DK = 9, NR = 10), na_values = c(9, 10)))
test$A
# <labelled_spss<integer>[10]>
#  [1]  1  2  3  4  5  6  7  8  9 10
# Missing values: 9, 10

# Labels:
#  value label
#      1  Good
#      8   Bad
#      9    DK
#     10    NR

write_sav(test, "test.sav")

testback <- read_sav("test.sav")
testback$A
# <labelled<double>[10]>
#  [1]  1  2  3  4  5  6  7  8  9 10

# Labels:
#  value label
#      1  Good
#      8   Bad
#      9    DK
#     10    NR
```

gorcha:
This PR fixes #596, adding write support for SPSS user missing values for integers as well as doubles.

This includes a minor change to the readstat SAV writing code - this change has been made upstream and will come through in the next readstat release (see WizardMac/ReadStat#251).

One thing to note is that we need to allow for either integer or real values in the na_range attribute if the main vector is either type, since we can't cast na_range to the vector type in labelled_spss().
Although na_values are cast to the type of the vector in labelled_spss(), na_range has to allow numeric vectors so that Inf/-Inf can be represented (the equivalent of SPSS HI and LO placeholders).

### DIFF
--- a/src/DfWriter.cpp
+++ b/src/DfWriter.cpp
@@ -269,8 +269,6 @@ public:
 
     if (Rf_inherits(x, "haven_labelled_spss")) {
       SEXP na_range = x.attr("na_range");
-      // We have to allow for both REAL and INTEGER here, since na_range cannot
-      // be cast to the vector type
       if (TYPEOF(na_range) == REALSXP && Rf_length(na_range) == 2) {
         readstat_variable_add_missing_double_range(var, REAL(na_range)[0], REAL(na_range)[1]);
       } else if (TYPEOF(na_range) == INTSXP && Rf_length(na_range) == 2) {
@@ -317,8 +315,6 @@ public:
 
     if (Rf_inherits(x, "haven_labelled_spss")) {
       SEXP na_range = x.attr("na_range");
-      // We have to allow for both REAL and INTEGER here, since na_range cannot
-      // be cast to the vector type
       if (TYPEOF(na_range) == REALSXP && Rf_length(na_range) == 2) {
         readstat_variable_add_missing_double_range(var, REAL(na_range)[0], REAL(na_range)[1]);
       } else if (TYPEOF(na_range) == INTSXP && Rf_length(na_range) == 2) {
--- a/src/DfWriter.cpp
+++ b/src/DfWriter.cpp
@@ -266,6 +266,25 @@ public:
     readstat_variable_set_label_set(var, labelSet);
     readstat_variable_set_measure(var, measureType(x));
     readstat_variable_set_display_width(var, displayWidth(x));
+
+    if (Rf_inherits(x, "haven_labelled_spss")) {
+      SEXP na_range = x.attr("na_range");
+      // We have to allow for both REAL and INTEGER here, since na_range cannot
+      // be cast to the vector type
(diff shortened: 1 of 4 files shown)


src/DfWriter.cpp
```diff
@@ -266,6 +266,25 @@ class Writer {
     readstat_variable_set_label_set(var, labelSet);
     readstat_variable_set_measure(var, measureType(x));
     readstat_variable_set_display_width(var, displayWidth(x));
+
+    if (Rf_inherits(x, "haven_labelled_spss")) {
+      SEXP na_range = x.attr("na_range");
+      // We have to allow for both REAL and INTEGER here, since na_range cannot
```
hadley, line 272: I _think_ you can omit this comment? It doesn't surprise me that you have to branch on the data type here, and I find the comment raises more questions than it answers (i.e. why can't you cast `na_range`? what is the "vector" type?)

hadley (approved):
A couple of small suggestions; feel free to merge once you've resolved them.

src/DfWriter.cpp
gorcha: Yeah fair call, it is a bit unnecessary/confusing here.

I feel like it's a bit of info that should be noted somewhere for posterity (mostly because my first instinct was to cast `na_range` to the labelled vector type before I realised it broke the tests :stuck_out_tongue:), maybe a comment in the `labelled_spss()` function near the `na_values` cast would be better?
https://github.com/gorcha/haven/blob/65f898276ad7eec9118892f463e7ba8f70542369/R/labelled_spss.R#L27-L29

Would you usually document these kinds of internal class assumptions, or is it better to let the code do the talking?

src/DfWriter.cpp
```diff
@@ -266,6 +266,25 @@ class Writer {
...
+      // be cast to the vector type
+      if (TYPEOF(na_range) == REALSXP && Rf_length(na_range) == 2) {
+        readstat_variable_add_missing_double_range(var, REAL(na_range)[0], REAL(na_range)[1]);
+      } else if (TYPEOF(na_range) == INTSXP && Rf_length(na_range) == 2) {
+        readstat_variable_add_missing_double_range(var, INTEGER(na_range)[0], INTEGER(na_range)[1]);
```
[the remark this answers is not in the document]
gorcha: Done, thanks!

src/DfWriter.cpp
hadley: If it clearly broke a test, I'd often just rely on that to reassure me that I wouldn't make the same mistake again in the future. I have a bad habit of ignoring my past comments because they don't make sense to me 😄

gorcha: Haha good call! That sounds good to me, I'll just remove the comment then merge the change.

Thanks!

Issue #624, opened 2021-08-01
READ_SAS() doesn't read label when variable class is (“POSIXct” “POSIXt”)


AntonShemiakin:
Hi,

currently I’m working with haven.read_sas() function. I’ve probably found a bug when I have variable with label in dataset which has class is (“POSIXct” “POSIXlt”).

Usually variables have two attributes: label and class. But in case when class is (“POSIXct” “POSIXt”) there is three attributes: timezone, label, class. 

I resolved this issue by manually change variable attributes order so that label always will be on the first place.

In my opinion, it can be useful to adjust on the haven.read_sas() function’s code.

Here is my MWE / MRE:

```r
---
author: ashemiak
date: 2021-08-01
output: "reprex::reprex\\_document"
title: grave-cobra\_reprex.R
---

``` r
library(data.table)
#> Warning: package 'data.table' was built under R version 4.0.5
library(haven)
#> Warning: package 'haven' was built under R version 4.0.5
library(reprex)
#> Warning: package 'reprex' was built under R version 4.0.5

aaa <- data.table(datetime = c(1550053200, 1553162400, 1556616900, 1561716300, 1570010100, 1570528620, 1574763600, 1577440800, 1580722800, 1542191700, 1543831080),
                  date = c(17940, 17976, 18016, 18075, 18171, 18177, 18226, 18257, 18295, 17849, 17868))

aaa$datetime <- as.POSIXct(aaa$datetime, origin = '1970-01-01', tz = 'GMT')
aaa$date <- as.Date(aaa$date, tryFormats = '%Y-%m-%d', origin = '1970-01-01')

attributes(aaa$datetime)$label <- 'example datetime'
attributes(aaa$date)$label <- 'example date'

write_sas(aaa, 'C:/Users/ashemiak/Desktop/bbb.sas7bdat')
get_bbb <- read_sas('C:/Users/ashemiak/Desktop/bbb.sas7bdat')
```

<sup>Created on 2021-08-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

<details>
```
Here is initial dataset AAA:

![aaa](https://user-images.githubusercontent.com/87488380/127767661-c65eba95-1bda-48a6-991a-df9834d7baa8.JPG)

Here is dataset GET_BBB after read_sas() execution:

![image](https://user-images.githubusercontent.com/87488380/127767702-61056aa6-8003-4d8e-a0e2-a351fcfcc8c1.png)

[...]

PR #631, opened 2021-08-14
Preserve attributes in output timezone conversion

gorcha:
The shared function that converts timezones to UTC in the `write_*` functions uses `as.POSIXct`, which drops all attributes (variable label etc.) (#624).

This PR preserves the attributes. I've also expanded the date time tests slightly so the datetime roundtrip performs the same tests for all file types.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -12,7 +12,11 @@ force_utc <- function(x) {
   if (identical(attr(x, "tzone"), "UTC")) {
     x
   } else {
-    as.POSIXct(format(x, usetz = FALSE), tz = "UTC", format = "%Y-%m-%d %H:%M:%S")
+    x_attr <- attributes(x)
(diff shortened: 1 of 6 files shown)


R/utils.R
```diff
@@ -12,7 +12,11 @@ force_utc <- function(x) {
   if (identical(attr(x, "tzone"), "UTC")) {
     x
   } else {
-    as.POSIXct(format(x, usetz = FALSE), tz = "UTC", format = "%Y-%m-%d %H:%M:%S")
+    x_attr <- attributes(x)
```
hadley, line 15: I think I'd add a couple of tests specifically about `force_utc()`, not just the user facing behaviour. Here I'd check that the correct attributes were preserved (i.e. not `class`)

gorcha: Shall do, thanks!

hadley (approved):

PR #633, opened 2021-08-28
read_sas_chunked

Shaunson26:
Attached is a read_sas_chunked function. It makes use of read_sas(skip, nmax) and a callback function to read a sas file by chunks. Much like readr::read_csv_chunked but without the R6 class. I've included the function + documentation, a small test using the sas7bdat files in the test folder. I ran devtools::document to generate the help documentation and add to the NAMESPACE.

Long time listener, first time caller. Please let me know if I've missed anything in the process of making the function and/or opening the pull request.

Should the function live in the haven-sas.R file? It is currently separate, but I included the test within the test-haven-sas.R
Should the test use another data set? Currently chunk_size = 1 as the test sas7bdat is very small


Shaunson26:
A few extra files complied when I tested this build on windows. I didn't include in the commit, and they've since disappeared from from the directory. These errors look related to spss functions.

gorcha:
Thanks for the contribution! Unfortunately, I don't think this is a good fit for the package. haven exposes the chunked API to enable this kind of programming, but maintaining a set of chunked reader functions is out of our scope.

Even though I'm closing this PR, I think this is a great example case for using the chunked API - would you be interested in putting together a short chunked API vignette?

Shaunson26:
What is the best approach provide the vignette? Through a PR, or just pass you on an Rmd?

gorcha:
Through a PR is best. Thank you!

Issue #634, opened 2021-09-03
write_dta POSIXct dates silently dropped


brianstamper:
I had a `data.frame` with dates stored as `POSIXct`, and I wrote that as a Stata file using `write_dta`. In the resulting `.dta` file all of the dates were blank. I found that this doesn't happen to `Date` types, so I could convert them and it would work, but figured this warranted reporting.

EDIT to show this affects POSIXlt as well, but now I notice these are also being encoded as POSIXct, I don't know why.
```R
library(haven)

test_data_out <- data.frame(my_POSIXct_date = as.POSIXct(c('2016-06-13',
                                                           '2015-08-17',
                                                           '2014-11-03',
                                                           '2012-10-01',
                                                           '2016-11-08')),
                            my_POSIXlt_date = as.POSIXlt(c('2016-06-13',
                                                           '2015-08-17',
                                                           '2014-11-03',
                                                           '2012-10-01',
                                                           '2016-11-08')),
                            my_Date_date = as.Date(c('2016-06-13',
                                                     '2015-08-17',
                                                     '2014-11-03',
                                                     '2012-10-01',
                                                     '2016-11-08')))

```
```R
> str(test_data_out)
'data.frame':	5 obs. of  3 variables:
 $ my_POSIXct_date: POSIXct, format: "2016-06-13" ...
 $ my_POSIXlt_date: POSIXct, format: "2016-06-13" ...
 $ my_Date_date   : Date, format: "2016-06-13" ...
```

```R
write_dta(test_data_out, 'test_data.dta')

test_data_in <- read_dta('test_data.dta')
```
```R
> str(test_data_in)
tibble [5 × 3] (S3: tbl_df/tbl/data.frame)
 $ my_POSIXct_date: POSIXct[1:5], format: NA ...
 $ my_POSIXlt_date: POSIXct[1:5], format: NA ...
 $ my_Date_date   : Date[1:5], format: "2016-06-13" ...
> test_data_in$my_POSIXct_date
[1] NA NA NA NA NA
> test_data_in$my_POSIXlt_date
[1] NA NA NA NA NA
```

brianstamper:
BTW although my example doesn't demonstrate this, the dates are in fact blank in Stata, so I do think it is on the `write_dta` side and not `read_dta`.

brianstamper:
The workaround I'm using with `lubridate::is.POSIXt` and `lubridate::as_date`:

```R
test_data_out %>% 
[...]

PR #636, opened 2021-09-08
Allow for POSIX dates with no time component (e.g. "2010-01-01") in output timezone conversion

gorcha:
The function that converts timezones to UTC in the `write_*` functions uses `as.POSIXct` with the format string `"%Y-%m-%d %H:%M:%S"`, which forces POSIX dates without a time component to `NA` (#634).

This PR allows for both date time (`"%Y-%m-%d %H:%M:%S"`) and date (`"%Y-%m-%d"`) format strings.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -13,8 +13,8 @@ force_utc <- function(x) {
     x
   } else {
     x_attr <- attributes(x)
-    x <- as.POSIXct(format(x, usetz = FALSE, format = "%Y-%m-%d %H:%M:%S"),
-                    tz = "UTC", format = "%Y-%m-%d %H:%M:%S")
+    x <- format(x, usetz = FALSE, format = "%Y-%m-%d %H:%M:%S")
+    x <- as.POSIXct(x, tz = "UTC", format = "%Y-%m-%d %H:%M:%S")
     attr_miss <- setdiff(names(x_attr), c(names(attributes(x)), "names"))
     attributes(x)[attr_miss] <- x_attr[attr_miss]
     x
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -13,7 +13,7 @@ force_utc <- function(x) {
     x
   } else {
     x_attr <- attributes(x)
-    x <- as.POSIXct(format(x, usetz = FALSE), tz = "UTC", format = "%Y-%m-%d %H:%M:%S")
+    x <- as.POSIXct(format(x, usetz = FALSE), tz = "UTC", tryFormats = c("%Y-%m-%d %H:%M:%S", "%Y-%m-%d"))
```
hadley, line 16: I think it would be slightly cleaner to set the date time format in `format()`, so we can just parse a single format.

gorcha: Good call, changed!

R/utils.R
```diff
@@ -13,7 +13,8 @@ force_utc <- function(x) {
     x
   } else {
     x_attr <- attributes(x)
-    x <- as.POSIXct(format(x, usetz = FALSE), tz = "UTC", format = "%Y-%m-%d %H:%M:%S")
+    x <- as.POSIXct(format(x, usetz = FALSE, format = "%Y-%m-%d %H:%M:%S"),
```
hadley, line 16: Now might be worth splitting into two steps? Convert to string, and then parse back into dates. That would avoid the awkward line wrapping.

hadley (approved):
One last code style comment — feel free to merge after considering it (i.e. it's up to you, just please read and consider it 😀); I don't need to look at it again.

Issue #638, opened 2021-09-15
Add `user_na` option to `zap_labels()`


gorcha:
Currently `zap_labels()` converts user missing values in labelled_spss objects to `NA`.

This should be controllable by the user. I think adding a `user_na` option like `read_sav()` is the best approach, where `FALSE` (default) converts to `NA` and `TRUE` keeps them as valid values.

``` r
library(haven)
x <- labelled_spss(1:3, na_values = 1)
# Current behaviour
zap_labels(x)
#> [1] NA  2  3

# Desired behaviour
zap_labels(x, user_na = TRUE)
#> [1] 1  2  3
```

<sup>Created on 2021-09-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

LuiNov:
Hey, 

I have an issue reading a dta file with several missings. Numerics are -5:0 and labels are ("Missing:Other", "Not askey") etc. . How can I specify those to the function so I get the clean dataset ?????

I think it doesn't fit into this issue but I didn't know elsewhere to comment.... Thanks a lot :-)

gorcha:
Hi @LuiNov - I've just moved this to its own issue at #645, shall reply over there.

PR #640, opened 2021-09-18
Add `user_na` argument to `zap_labels()`

gorcha:
This PR adds a `user_na` argument to `zap_labels()` to control treatment of user-defined missing values for `labelled_spss()` vectors (#638).

### DIFF
--- a/R/zap_labels.R
+++ b/R/zap_labels.R
@@ -5,9 +5,14 @@
 #' want to simply drop all `labels` from a data frame.
 #'
 #' Zapping labels from [labelled_spss()] also removes user-defined missing
-#' values, replacing with standard `NA`s.
+#' values by default, replacing with standard `NA`s. Use the `user_na` argument
+#' to override this behaviour.
 #'
 #' @param x A vector or data frame
+#' @param user_na If `FALSE`, the default, `zap_labels()` will convert
+#'   [labelled_spss()] user-defined missing values to `NA`. If `TRUE` they
+#'   will be treated like normal values.
+#' @param ... Other arguments passed down to method.
 #' @family zappers
 #' @seealso [zap_label()] to remove variable labels.
 #' @export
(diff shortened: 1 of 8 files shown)


R/zap_labels.R
```diff
@@ -20,30 +25,36 @@
...
+#' # Keep the user defined missing values
+#' zap_labels(x2, user_na = TRUE)
+#'
 #' # zap_labels also works with data frames
 #' df <- tibble::tibble(x1, x2)
 #' df
 #' zap_labels(df)
-zap_labels <- function(x) {
+zap_labels <- function(x, ...) {
   UseMethod("zap_labels")
```
hadley, line 36: When you add `...` to an argument it's good practice to use an ellipsis function (e.g. `ellipsis::check_dots_used()`) to check that they're used appropriately.

R/zap_labels.R
```diff
@@ -54,8 +65,9 @@ zap_labels.haven_labelled_spss <- function(x) {
 }
 
 
+#' @rdname zap_labels
 #' @export
-zap_labels.data.frame <- function(x) {
-  x[] <- lapply(x, zap_labels)
+zap_labels.data.frame <- function(x, user_na = FALSE, ...) {
+  x[] <- lapply(x, zap_labels, user_na = user_na, ...)
```
hadley, line 71: I don't think you need `user_na` here, just `...`

R/zap_labels.R
```diff
@@ -20,30 +25,36 @@
...
   class(x) <- NULL
 
   x
 }
 
+#' @rdname zap_labels
 #' @export
-zap_labels.haven_labelled_spss <- function(x) {
-  x[is.na(x)] <- NA
+zap_labels.haven_labelled_spss <- function(x, user_na = FALSE, ...) {
```
hadley, line 54: `user_na` should go after `...` — https://design.tidyverse.org/dots-position.html

gorcha: Done!

R/zap_labels.R
gorcha: Good call, updated!

R/zap_labels.R
gorcha: I'm not sure about this one, because we probably want everything except `zap_labels.haven_labelled_spss()` to silently eat `user_na` to avoid unnecessary warnings when running over a data frame with non-`labelled_spss()` objects. For e.g., when adding `ellipsis::check_dots_used()` to the generic:
``` r
library(haven)

df <- tibble::tibble(
  x = 1:10,
  y = labelled_spss(10:1, c("good" = 1), na_values = 10)
)
zap_labels(df, user_na = TRUE)
#> Warning: 1 components of `...` were not used.
#> 
#> We detected these problematic arguments:
#> * `user_na`
#> 
#> Did you misspecify an argument?
#> # A tibble: 10 × 2
#>        x     y
#>    <int> <int>
#>  1     1    10
#>  2     2     9
#>  3     3     8
#>  4     4     7
#>  5     5     6
#>  6     6     5
#>  7     7     4
#>  8     8     3
#>  9     9     2
#> 10    10     1
```

<sup>Created on 2021-09-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Putting `ellipsis::check_dots_used()` in `zap_labels.haven_labelled_spss()` instead wouldn't cover everything but would save users from misspelling `user_na` at least.

What would you do in this case?

hadley: Oh hmmm, good question.

hadley (approved):
Fine to leave `...` behaviour as is for now.

Issue #641, opened 2021-09-24
`write_sav()` does not check for duplicated variable names


sjkiss:
I am trying to export a data set as an sav file via `write_sav`. It would be convenient for me to have haven_labelled versions of variables as well as pure factor variables. I'm trying to export the data with the labelled versions of the variables with lower case names and with the factor version of the variables 

```r
#This creates my data frame with one labelled variable and a duplicate factor
test<-structure(list(francophone = structure(c(0, 0, 0, 0, 0, 0, 0, 0, 0, 0), labels = c(Francophone = 1, `Not Francophone` = 0), label = "Dichotomous variable, R is francophone", class = c("haven_labelled", "vctrs_vctr", "double")), Francophone = structure(c(1L, 1L, 1L,1L, 1L, 1L, 1L, 1L, 1L, 1L), .Label = c("Not Francophone", "Francophone"), label = "Dichotomous variable, R is francophone", class = "factor")), row.names = c(NA,-10L), class = c("tbl_df", "tbl", "data.frame"))

#Create a second factor but with a significantly different name
test$Franco<-test$Francophone
#Export the data-set and look.
test %>% 
  write_sav(., path="~/Desktop/test2.sav")

```
<img width="1159" alt="Screen Shot 2021-09-24 at 10 32 30 AM" src="https://user-images.githubusercontent.com/5245117/134693254-fd5c51a9-e9df-492d-9a19-f65aaf4577f2.png">

My desired outcome would be for `write_sav()` to treat variable names case sensitively

gorcha:
Hi!

This is an SPSS thing - SPSS is itself case insensitive, so there's no way to write "francophone" and "Francophone" as separate variables unfortunately.

Really this should throw an error when attempting to write the file in haven, so we really need to add variable name validation to `write_sav()`.

PR #643, opened 2021-09-28
added validation that column names are unique in export to SPSS

juansebastianl:
A fix to issue #641

### DIFF
--- a/R/haven-spss.R
+++ b/R/haven-spss.R
@@ -109,6 +109,18 @@ read_spss <- function(file, user_na = FALSE, col_select = NULL, skip = 0, n_max 
 validate_sav <- function(data) {
   stopifnot(is.data.frame(data))
 
+  #check unique column names 
+  col_names <- names(data)
+  unique_colnames <- unique(tolower(names(data)))
+  if(length(col_names) != length(unique_colnames)){
+    stop(
+      "SPSS requires unique column names\n",
+      "column(s) \"", paste(col_names[duplicated(tolower(col_names))], collapse = "\" \""), "\" not unique\n", 
+      "reminder: SPSS is not case sensitive"
+    )
+
+  }
+
   # Check factor lengths
   level_lengths <- vapply(data, max_level_length, integer(1))
 


juansebastianl:
I don't know how to add tests here, but this PR is something that can easily be unit tested.

R/haven-spss.R
```diff
@@ -109,6 +109,18 @@ read_spss <- function(file, user_na = FALSE, col_select = NULL, skip = 0, n_max
 validate_sav <- function(data) {
   stopifnot(is.data.frame(data))
 
+  #check unique column names 
+  col_names <- names(data)
+  unique_colnames <- unique(tolower(names(data)))
+  if(length(col_names) != length(unique_colnames)){
+    stop(
```
gorcha, line 116: Error messages should have `call. = FALSE` to  [exclude call information](https://style.tidyverse.org/error-messages.html#problem-statement).

R/haven-spss.R
```diff
@@ -109,6 +109,18 @@ read_spss <- function(file, user_na = FALSE, col_select = NULL, skip = 0, n_max
...
+      "SPSS requires unique column names\n",
+      "column(s) \"", paste(col_names[duplicated(tolower(col_names))], collapse = "\" \""), "\" not unique\n",
```
gorcha, line 118: Variable names should be [surrounded by backticks](https://style.tidyverse.org/error-messages.html#punctuation) instead of double quotes - haven has a handy internal `var_names()` helper for this, see `bad_lengths` below for an example.

R/haven-spss.R
```diff
@@ -109,6 +109,18 @@ read_spss <- function(file, user_na = FALSE, col_select = NULL, skip = 0, n_max
 validate_sav <- function(data) {
   stopifnot(is.data.frame(data))
 
+  #check unique column names 
+  col_names <- names(data)
+  unique_colnames <- unique(tolower(names(data)))
```
gorcha, line 114: Not essential, just a suggestion - a cleaner way to do this might be to create a `dupe_vars <- duplicated(tolower(names(data)))` to avoid having to perform the duplicate check in both the if statement and the column name filtering in the error message.

gorcha (changes requested):
Thanks for the contribution! I've made a few comments, just style related stuff.

Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Regarding a unit test - a good reference for adding a test covering this functionality is the Stata variable name checks in "test-haven-stata.R" (this tests the checks for variable names that are too long or include invalid characters, so it's a little different but the same general idea):
https://github.com/tidyverse/haven/blob/eb5e8c9fe2b56b269079942690452c53993b5ac6/tests/testthat/test-haven-stata.R#L217-L230

R/haven-spss.R
```diff
@@ -109,6 +109,18 @@ read_spss <- function(file, user_na = FALSE, col_select = NULL, skip = 0, n_max
...
+  if(length(col_names) != length(unique_colnames)){
```
gorcha, line 115: Place a [space before and after `()`](https://style.tidyverse.org/syntax.html#spacing) when used with `if`, `for`, or `while`.

gorcha:
Hi @juansebastianl, I've opened a new PR with a little clean up of this check and some tests added alongside a couple of other new variable name validation rules in #660, so closing this off.

Thanks again for the contribution, very much appreciated! :slightly_smiling_face:

PR #644, opened 2021-09-29
Update GitHub Actions workflows

gorcha:
This PR updates to the most recent GitHub actions templates, prompted by [deprecation of the Ubuntu 16.04 runners](https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/).

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -25,8 +25,6 @@ jobs:
           - {os: macOS-latest,   r: 'release'}
 
           - {os: windows-latest, r: 'release'}
-          # Use 3.6 to trigger usage of RTools35
-          - {os: windows-latest, r: '3.6'}
 
           # Use older ubuntu to maximise backward compatibility
           - {os: ubuntu-18.04,   r: 'devel', http-user-agent: 'release'}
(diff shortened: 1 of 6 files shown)


hadley (approved):
Thanks for taking care of this!

gorcha:
No worries! Also prepares us for #635

There are a couple of build failures at the moment - the ubuntu-18.04 devel build looks like it's just a race condition (waiting for packages to be compiled on RSPM), but there's also an odd Windows compile failure (of course!):
```
 /usr/bin/sh: -c: line 9: syntax error: unexpected end of file
make: *** [C:/R/share/make/winshlib.mk:13: haven.dll] Error 1
ERROR: compilation failed for package 'haven'
```

There aren't any errors in the actual compilation so it looks like it might be an issue with the old Rtools35 build tooling.
Just wondering if you've seen anything like this pop up before?

hadley:
Given that Rtools35 failure isn't causing any noticeable problems (i.e. we're not seeing bug reports from windows users), I think we can just drop it.

gorcha:
OK cool, shall do :slightly_smiling_face: Thanks!

Issue #661, opened 2022-02-28
Use `abort()` instead of `stop()` for error messaging


gorcha:
In line with current tidyverse best practices we should use `rlang::abort()` for error messaging instead of `stop()`.

PR #673, opened 2022-04-06
Tidyverse upkeep

gorcha:
This PR makes a bunch of tidyverse upkeep changes, closing #672 and #661.

The [condition formatting](https://rlang.r-lib.org/reference/topic-condition-formatting.html) write-up in rlang recommends using `cli::cli_abort()` instead of `rlang::abort()`, but it's relatively new so I'm a bit unsure about replacing all the error messaging with it yet (it would move cli from suggests to imports and need >=3.0.0) - what do you think?


hadley:
It's fine to move directly to `cli_abort()` if you want to. It makes it much easier to construct error messages since you can use glue.

gorcha:
OK cool, have done! I think we're ready for release once this is merged.

hadley (approved):
Looks great, thanks for working on this!

Do you want to start the release process too? (i.e. call `use_release_issue()` and then start on the bullets?)

gorcha:
No worries at all :slightly_smiling_face:

And yep, shall do!

PR #686, opened 2022-06-23
Allow functions in `.name_repair`

### LINKED ISSUE #684: `.name_repair` does not accept a function  in `read_`
It only works with the string

I tested with `.dta` and `.sav` and I get the same error message.
```r
read_dta("file.dta", .name_repair = make.names)
#> Error expected string vector of length 1.
```
In the doc., it says that it accepts rlang functions.

gorcha:
This PR fixes #684.

I've modified the `.name_repair` argument to allow an arbitrary SEXP instead of a `std::string`, allowing users to pass a function in `.name_repair` to `tibble::as_tibble()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -43,6 +43,6 @@ Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.0
 SystemRequirements: GNU make, C++11, zlib: zlib1g-dev (deb), zlib-devel
     (rpm)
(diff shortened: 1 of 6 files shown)


olivroy:
Just a comment / question. WIll this work
```r
  df <- data.frame(1:3, 1:3)
  colnames(df) <- c("id", "id")
  path <- tempfile()
  write_sas(df, path)

read_sas(path, .name_repair = "universal", col_select = c(id...1))
```
Just like in readr.

gorcha:
Hi @olivroy,

This PR just updates the way the `.name_repair` argument is passed.
There's a bit of an odd interaction between `col_select` and `.name_repair` at the moment, which is causing the issue you're seeing, but it's not really related. I'll open a new issue for this.

Issue #702, opened 2022-12-14
Data quality issues associated with decision to force UTC on date/time objects (plus related bug)


jmobrien:
Wanted to bring up some concerns associated with the changes to force implied UTC on all date/time variables prior to exporting (associated with #555).   

I get that the best choice isn't obvious.  As discussed there and later in #692, since the other suites all lack a notion of time zone, while the date/time variables all "officially" represent UTC (or similar, for SPSS), even looking at their own docs it's encouraged to just use them to display as expected, so the package is following that expectation.   

However, even if it's reasonable to consider users display expectations for outside suites, I think users likely have a deeper, implicit expectation that display should take a back seat to data quality and accuracy.  In the case of dates and times, an important accuracy concern is a consistent metric *between* time points in data. Unfortunately, thanks to the complexities of things daylight savings time, the current approach might be falling short.

Below is a reprex that resembles an actual scenario from my team's data collection: multi-day intensive data collection for an individual (including biological sampling, for which things like circadian rhythms--and thus accurate timestamps--matter).  We also happen to have complex data use/publication demands that means that while we mostly work in R, we also have to support Stata datasets.  

In this case, a 10-day collection window happens to fall over a DST transition day.  So, despite consistent collection "by the clock" each day, there's actually some variation in sampling times that is important to take into account.

Pre-#555, all timestamps would be shown in UTC once sent to Stata, regardless of collection locale.  That was important for us to document, but users at least knew that data reliably measured the passage of time from one collection to the next.  But now that all timestamps are treated as if they were always UTC to start with, our Stata users and R users are actually getting different timestamp data:

``` r
require(haven, quietly = TRUE)
require(tidyverse, quietly = TRUE)

# Function because diff.POSIXt can't specify units:
alt_diff <- 
  function(x){
    difftime(x[-1], x[-length(x)], units = "hours") |> 
      append(as.difftime(NA_character_, units="hours"), after = 0)
  }

# Data from someone sampled 10 sequential days at 8:00 AM,
# where collection happened to span an end-of-daylight-savings transition:
df <- 
  tibble(
    session = 
      1:10,
    time = 
      glue::glue("2022-11-{1:10} 08:00:00") |> 
      as.POSIXct(tz = "America/Chicago"),
    difftime_orig = 
      alt_diff(time)
  )

# Time data + relationship between time points is accurate:
# - one day-over-day time difference 25 hours due to DST ending
# - subsequent sessions stay at that later time (24 hours past previous)

df
[...]

PR #703, opened 2022-12-14
Original data return from write_* functions no longer has UTC-adjusted date/times

jmobrien:
Addresses secondary issue mentioned in #702.  

Noting ahead of time: it's possible changes made here might not be necessary/a good fit if other changes are made in response to #702, and if so this PR can be ignored.

Relatedly, if other #702-related changes (e.g., reverting to older behavior, and/or adding a new selector argument for how conversion is done) are both desired and easier if just added to this PR, I'd be glad to do that.


R/haven-sas.R
```diff
@@ -146,8 +149,6 @@ write_xpt <- function(data, path, version = 8, name = NULL, label = attr(data, "
 
 validate_sas <- function(data) {
   stopifnot(is.data.frame(data))
-
-  adjust_tz(data)
```
gorcha, line 150: Now that it's not adjusting the timezone I'd prefer that it just returns `invisible(data)` for so there's a consistent return value for all of the `validate_*()` functions

R/haven-spss.R
```diff
@@ -165,8 +166,6 @@ validate_sav <- function(data, call = caller_env()) {
       call = call
     )
   }
-
-  adjust_tz(data)
```
gorcha, line 169: Now that it's not adjusting the timezone I'd prefer that it just returns `invisible(data)` for so there's a consistent return value for all of the `validate_*()` functions

R/haven-stata.R
```diff
@@ -147,7 +151,6 @@ validate_dta <- function(data, version, call = caller_env()) {
     )
   }
 
-  adjust_tz(data)
```
gorcha, line 150: Now that it's not adjusting the timezone I'd prefer that it just returns `invisible(data)` for so there's a consistent return value for all of the `validate_*()` functions

gorcha (commented):
Hi @jmobrien, thanks for the PR! All looking good, just one thing - I'd prefer all the `validate_*()` functions to return `invisible(data)` so they have a consistent return value.

R/haven-sas.R
jmobrien: Done, in [41f21fe](https://github.com/tidyverse/haven/pull/703/commits/41f21fe0eb5bea54b02206610dcdffedf2ae14b7)

R/haven-spss.R
jmobrien: Done, in [41f21fe](https://github.com/tidyverse/haven/pull/703/commits/41f21fe0eb5bea54b02206610dcdffedf2ae14b7)

R/haven-stata.R
jmobrien: Done, in [41f21fe](https://github.com/tidyverse/haven/pull/703/commits/41f21fe0eb5bea54b02206610dcdffedf2ae14b7)

PR #714, opened 2023-02-21
Add `convert_utc` argument to writer functions

### LINKED ISSUE #702: Data quality issues associated with decision to force UTC on date/time objects (plus related bug)

gorcha:
Close #702.

@jmobrien do you have any feedback on the docs? I'm not entirely sure that this makes sense :sweat_smile: 

> If `TRUE` (the default) date times are converted to the equivalent UTC value and timezone is ignored, so they will appear the same in R and Stata/SPSS/SAS. If `FALSE`, date time variables are written as the corresponding UTC value.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,13 @@
 # haven (development version)
 
+* `write_*()` functions gain a new `convert_utc` argument to control timezone
+  conversion (#702) for date time variables. If `TRUE` (the default) date times
+  are converted to the equivalent UTC value and timezone is ignored, so they
+  will appear the same in R and Stata/SPSS/SAS. If `FALSE`, date time variables
+  are written as the corresponding UTC value.
+
 * Fixed issue in `write_*()` functions where invisible return of input data 
-  frame included unintended alteration of date/time variables. (@jmobrien, #702)
+  frame included unintended alteration of date time variables. (@jmobrien, #702)
 
 * The experimental `write_sas()` function has been deprecated (#224). The
   sas7bdat file format is complex and undocumented, and as such writing SAS
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,11 +7,20 @@
   * Ignore invalid SAV timestamp strings (#683).
   * Fix compiler warnings (#707).
 
-* `write_*()` functions gain a new `convert_utc` argument to control timezone
-  conversion (#702) for date time variables. If `TRUE` (the default) date times
-  are converted to the equivalent UTC value and timezone is ignored, so they
-  will appear the same in R and Stata/SPSS/SAS. If `FALSE`, date time variables
-  are written as the corresponding UTC value.
+* `write_*()` functions gain a new `adjust_tz` argument to allow more control
+  over time zone conversion for date-time variables (#702). Thanks to @jmobrien
+  for the detailed issue and feedback.
+  
+  Stata, SPSS and SAS do not have a concept of time zone. Since haven 2.4.0
+  date-time values in non-UTC time zones are implicitly converted when writing
+  to ensure the time displayed in Stata/SPSS/SAS will match the time displayed
+  to the user in R (see #555). This is the behaviour when `adjust_tz = TRUE`
+  (the default). Although this is in line with general user expectations it can
+  cause issues when the time zone is important, for e.g. when looking at
+  differences between time points, since the underlying numeric data is changed
+  to preserve the displayed time. Use `adjust_tz = FALSE` to write the time as
+  the corresponding UTC value, which will appear different to the user but
+  preserves the underlying numeric data.
 
 * Fixed issue in `write_*()` functions where invisible return of input data 
   frame included unintended alteration of date time variables. (@jmobrien, #702)
--- a/R/haven-sas.R
+++ b/R/haven-sas.R
@@ -134,11 +134,19 @@ read_xpt <- function(file, col_select = NULL, skip = 0, n_max = Inf, .name_repai
 #'
 #'   Note that although SAS itself supports dataset labels up to 256 characters
 #'   long, dataset labels in SAS transport files must be <= 40 characters.
-#' @param convert_utc If `TRUE` (the default), date times are converted to
-#'   the equivalent UTC value and timezone is ignored, so they will appear the
-#'   same in R and Stata/SPSS/SAS. If `FALSE`, date time variables are written
-#'   as the corresponding UTC value.
-write_xpt <- function(data, path, version = 8, name = NULL, label = attr(data, "label"), convert_utc = TRUE) {
+#' @param adjust_tz Stata, SPSS and SAS do not have a concept of time zone,
+#'   and all [date-time] variables are treated as UTC. `adjust_tz` controls
+#'   how the timezone of date-time values is treated when writing.
+#'
+#'   * If `TRUE` (the default) the timezone of date-time values is ignored, and
+#'   they will display the same in R and Stata/SPSS/SAS, e.g.
+#'   `"2010-01-01 09:00:00 NZDT"` will be written as `"2010-01-01 09:00:00"`.
+#'   Note that this changes the underlying numeric data, so use caution if
+#'   preserving between-time-point differences is critical.
+#'   * If `FALSE`, date-time values are written as the corresponding UTC value,
+#'   e.g. `"2010-01-01 09:00:00 NZDT"` will be written as
+#'   `"2009-12-31 20:00:00"`.
+write_xpt <- function(data, path, version = 8, name = NULL, label = attr(data, "label"), adjust_tz = TRUE) {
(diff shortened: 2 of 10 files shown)


R/haven-sas.R
```diff
@@ -134,7 +134,11 @@ read_xpt <- function(file, col_select = NULL, skip = 0, n_max = Inf, .name_repai
 #'
 #'   Note that although SAS itself supports dataset labels up to 256 characters
 #'   long, dataset labels in SAS transport files must be <= 40 characters.
-write_xpt <- function(data, path, version = 8, name = NULL, label = attr(data, "label")) {
+#' @param convert_utc If `TRUE` (the default), date times are converted to
```
hadley, line 137: I wonder if this would be slightly easier to understand if it was `convert_tz = c("utc", "current")` or similar?

NEWS.md
```diff
@@ -7,8 +7,14 @@
   * Ignore invalid SAV timestamp strings (#683).
   * Fix compiler warnings (#707).
 
+* `write_*()` functions gain a new `convert_utc` argument to control timezone
+  conversion (#702) for date time variables. If `TRUE` (the default) date times
+  are converted to the equivalent UTC value and timezone is ignored, so they
+  will appear the same in R and Stata/SPSS/SAS. If `FALSE`, date time variables
+  are written as the corresponding UTC value.
```
hadley, line 14: What does "corresponding UTC value" mean? I expected to see something about local/current timezone?

R/haven-sas.R
gorcha: I thinks it's clearer as a logical argument since it's essentially a "conversion" vs "no conversion" choice, and the main issue is that the explanation in the docs is a bit silly at the moment

NEWS.md
gorcha: Yeah I'm not super happy with the explanation here, trying to explain it was breaking my brain a bit!

What about:

> Stata, SPSS and SAS do not have a concept of time zone, and all date time variables are treated as UTC. `convert_utc` controls how the timezone of date time values is treated when writing.
>
> * If `TRUE` (the default) the timezone of date time values is ignored, and they will appear the same in R and Stata/SPSS/SAS, e.g. `"2010-01-01 09:00:00 NZDT"` will appear as `"2010-01-01 09:00:00"`.
> * If `FALSE`, date time variables are written as is, and will appear as the corresponding UTC value, e.g. `"2010-01-01 09:00:00 NZDT"` will appear as `"2009-12-31 20:00:00"`.

After writing that, I'm thinking maybe `adjust_tz` makes more sense as an argument name?

hadley: Yeah, that explanation and suggest argument name is much clearer 😄

gorcha:
Thanks @jmobrien, much appreciated! I've made a few tweaks based on your feedback.

Issue #766, opened 2024-11-15
Remove reference to write_sas function in read_sas help


szimmer:
Remove line from `read_sas()` documentation that states:

```
write_sas() returns the input data invisibly.
```

This isn't relevant and the function is deprecated which can lead to confusion.

PR #767, opened 2024-11-16
Remove write_sas mention from read_sas doc

szimmer:
This addresses #766 

Note that while updating documentation with `devtools::document()` a few other unrelated files were changed as documentation must not have been run since some other changes had been made.


gorcha:
Thanks!

PR #801, opened 2026-06-06
docs: document retained metadata for read_xpt/write_xpt (#719)

LeonidasZhak:
Addresses #719.

Documents the metadata retained by `read_xpt()` and used by `write_xpt()` for round-tripping:

- **`format.sas`**: Documents that the original SAS variable format is stored as `format.sas` and that `write_xpt()` reads it back to preserve the format during round-trips.
- **Value labels**: Clarifies that the SAS transport format does not support value labels (they are not retained by `read_xpt()` and silently ignored by `write_xpt()`).
- **Round-trip example**: Adds a runnable example showing that `label` and `format.sas` survive a `write_xpt()` → `read_xpt()` cycle.
- **Notes**: Documents the `notes` attribute for completeness.

This is a documentation-only change to `R/haven-sas.R`, `man/read_xpt.Rd`, and `NEWS.md`. Complements the existing `read_sas`/`read_dta`/`read_spss` metadata docs.

**Relation to #800**: PR #800 documents the `labelled()` class and value label workflow for `read_sas`/`read_dta`/`read_spss`. This PR focuses on the XPT-specific metadata that #800 does not cover. No overlap.

Checks run locally:
- `tools::checkRd("man/read_xpt.Rd")` — clean
- `R CMD build --no-build-vignettes .` — clean
- `_R_CHECK_FORCE_SUGGESTS_=false R CMD check --no-manual --ignore-vignettes --no-build-vignettes --no-tests haven_2.5.5.9000.tar.gz` — 1 pre-existing NOTE (`setNames` in NAMESPACE), unrelated to this change
- Round-trip example verified manually

PR #802, opened 2026-06-06
test: add test for format.sas trailing period stripping in XPT roundtrip

LeonidasZhak:
Adds a test that explicitly verifies the trailing period stripping behavior documented in PR #801.

The SAS transport format does not store trailing periods in format names (e.g., "DATE9." becomes "DATE9"). This test confirms that write_xpt() to read_xpt() correctly strips trailing periods from format.sas attributes.

This test complements the existing "can roundtrip format attribute" test, which only uses formats without trailing periods.

Checks run locally:
- devtools::test(filter = "haven-sas") - all 73 tests pass
- Full test suite: 1 pre-existing snapshot failure in test-labelled-pillar.R (Unicode encoding), unrelated to this change

### DIFF
--- a/tests/testthat/test-haven-sas.R
+++ b/tests/testthat/test-haven-sas.R
@@ -304,6 +304,21 @@ test_that("can roundtrip format attribute", {
   expect_identical(df, out)
 })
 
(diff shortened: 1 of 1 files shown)


LeonidasZhak:

